PR #1, opened 2020-10-22
Update to wdlTools 0.8.4

jdidion:
This fixes issues compiling some complex workflows that were rendered incorrectly by the code generator.

Note: I have been unable to reduce the issues with the customer workflow to a simple reproducible test case. I have tested locally that the customer's workflow compiles but I do not want to add the customer's workflow as a test case.

### DIFF
--- a/src/test/scala/dx/compiler/CompilerTest.scala
+++ b/src/test/scala/dx/compiler/CompilerTest.scala
@@ -957,6 +957,27 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
     trainsOutputVector.outputVars.size shouldBe 1
   }
 
+  it should "compile workflow with task imported by multiple paths" taggedAs NativeTest in {
+    val path = pathFromBasename("compiler", basename = "multi_import.wdl")
+    val args = path.toString :: cFlags
+    val retval = Main.compile(args.toVector)
+    retval shouldBe a[Success]
+  }
+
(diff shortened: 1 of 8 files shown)


src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -957,6 +957,27 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
+  }
+
+  it should "compile workflow with call to task having optional output" taggedAs NativeTest in {
+    val path = pathFromBasename("compiler", basename = "optional_call_output.wdl")
+    val args = path.toString :: cFlags
+    val retval = Main.compile(args.toVector)
+    retval shouldBe a[Success]
+  }
+
+  it should "compile myriad workflow" taggedAs NativeTest in {
```
commandlinegirl, line 974: It looks like this was left here by mistake.

commandlinegirl (changes requested):

src/test/scala/dx/compiler/CompilerTest.scala
jdidion: oops - removed it

commandlinegirl (approved):
LGTM, thanks!

commandlinegirl (changes requested):
Please when you get a chance check the integration tests as described here: https://jira.internal.dnanexus.com/browse/APPS-245

jdidion:
Abandoning this in favor of #7.

PR #2, opened 2020-10-22
APPS-235 Change streamAllFiles to streamFiles

jdidion:


commandlinegirl (commented):
Looks good, thanks for this addition! Can you also add an option to tests, like here: https://github.com/dnanexus/dxCompiler/blob/main/scripts/run_tests.py#L704. Is there an easy way to add a test for it to our standard integration test run (.scripts/run_tests.py)?

PR #5, opened 2020-10-30
APPS-249 Run integration tests from GHA

commandlinegirl:
This commit will make it easier to run integration tests for the WDL compiler. 

The changes include:
- fix unit tests which don't kick off - change master to main in the existing workflow
- add `--clean` option to the `run_tests.py` script that removes the build folder after running tests
- integration tests will run when the `integration` label is set on a PR. This is to limit running them to devs with write access to the repo (write access is needed to set labels) and also limit the number of runs so that we don't incur too much cost. 
- the test script will now fail with a non-zero exit code if any of the workflows fail.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,31 +1,34 @@
 name: Integration Test Workflow
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
-    types: [ labeled ]
+    branches:
+      - main
+    types:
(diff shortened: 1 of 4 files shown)


commandlinegirl:
Example of integration tests run: https://github.com/dnanexus/dxCompiler/actions/runs/342173254

jdidion (approved):
lgtm

PR #7, opened 2020-11-02
APPS-186, APPS-237, APPS-247, APPS-266 Fix handling of optional block inputs; fix "key no found" at runtime; fix two different callable definitions error; fix instance comparison; set wdlTools to 0.10.5; add dxCommon

jdidion:
Fixes: APPS-186, APPS-237, APPS-247, APPS-266

Previously, optional block inputs were ignored. This fix instead sets either a default or null value for missing optional block inputs, and throws an exception if there are any missing required block inputs.

This fix also addresses the situation where a call input references a field of the scatter variable, e.g.

```wdl
version 1.0

workflow scatter_with_eval {
  Array[String] names = ["Michael", "Lukas", "Martin", "Shelly", "Amy"]
  Array[Int] ages = [27, 9, 13, 67, 2]

  scatter (pair in zip(names, ages)) {
    String info = pair.left + "_" + pair.right
  }

  output {
    Array[String] result = info
  }
}
```

Previously, accessing `pair.left` would cause an exception. Now we have a special block input type `ComputedBlockInput` that is (currently only) used for the scatter variable, and signals that the variable is computed during evaluation of workflow inputs and is not required at compile time.

The bulk of the changed files are due to upgrading to wdlTools 0.10.5. In that version, common utilities were moved from wdlTools.util to dx.util in the new dxCommon package. The key changes are in four files:

* dx.core.languages.wdl
    * WdlBlock.scala
    * WdlUtils.scala
* dx.translator.wdl.CallableTranslator
* dx.executor.wdl.WdlWorkflowSupport

Edit by @commandlinegirl: It also updates the mechanism of comparing instance types to select the cheapest one for execution (APPS-266).


commandlinegirl:
@jdidion Would you review https://github.com/dnanexus/dxCompiler/pull/5 so we run it through the integration tests before merging?

src/main/scala/dx/executor/wdl/WdlWorkflowSupport.scala
```diff
@@ -914,13 +924,23 @@ case class WdlWorkflowSupport(workflow: TAT.Workflow,
   ): BlockContext[WdlBlock] = {
     val block: WdlBlock =
       Block.getSubBlockAt(WdlBlock.createBlocks(workflow.body), jobMeta.blockPath)
-    // Some of the inputs could be optional. If they are missing,
-    // add in a None value.
-    val irInputEnv: Map[String, (Type, Value)] = block.inputs.collect {
-      case blockInput: WdlBlockInput if jobInputs.contains(blockInput.name) =>
-        blockInput.name -> jobInputs(blockInput.name)
-    }.toMap
-    val inputEnv = WdlUtils.fromIR(irInputEnv, wdlTypeAliases)
```
commandlinegirl, line 923: The [fromIR()](https://github.com/dnanexus/dxCompiler/blob/2f2d26843c4a9a3d8f094fbf038a22d5ff512074/src/main/scala/dx/core/languages/wdl/WdlUtils.scala#L657) function can be removed

commandlinegirl (changes requested):
Minor comment but let's merge after #5. Thank you!

src/main/scala/dx/executor/wdl/WdlWorkflowSupport.scala
jdidion: APPS-247

commandlinegirl (approved):
LGTM, thanks!

jdidion:
I've migrated over changes from the PR updating to wdlTools 0.8.4 and further updated to 0.10.1, which includes another fix required to address this issue.

src/main/scala/dx/core/io/DxWorkerPaths.scala
```diff
@@ -15,8 +41,59 @@ import wdlTools.util.Logger
   * files created with stdlib calls like "write_lines".
   * @param rootDir the root directory - typically the user's home directory
   */
-case class DxWorkerPaths(rootDir: Path)
-    extends ExecPaths(rootDir, rootDir.resolve(DxWorkerPaths.TempDir)) {
+case class DxWorkerPaths(rootDir: Path) extends BaseEvalPaths with ExecPaths {
+  def getRootDir(ensureExists: Boolean = false): Path = {
+    getOrCreateDir("root", rootDir, ensureExists)
```
commandlinegirl, line 46: I cannot find where `getOrCreateDir` is defined. If rootDir is /home/dnanexus, is `getRootDir()` it going to get/create a directory at /home/dnanexus/root?

src/main/scala/dx/core/io/DxWorkerPaths.scala
```diff
@@ -15,8 +41,59 @@ import wdlTools.util.Logger
...
+  }
+
+  /**
+    * The file that has a copy of standard error.
+    */
+  def getStderrFile(ensureParentExists: Boolean = false): Path = {
+    getMetaDir(ensureParentExists).resolve(DxWorkerPaths.StderrFile)
+  }
+
+  def getCommandFile(ensureParentExists: Boolean = false): Path = {
```
commandlinegirl, line 82: Do we need them now if these metadirs are not used atm?

build.sbt
```diff
@@ -70,7 +70,8 @@ logLevel in assembly := Level.Info
 assemblyOutputPath in assembly := file("applet_resources/resources/dxWDL.jar")
 assemblyMergeStrategy in assembly := customMergeStrategy.value
 
-val wdlToolsVersion = "0.8.3"
+val dxCommonVersion = "0.2.0"
```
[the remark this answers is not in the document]
jdidion: https://github.com/dnanexus/dxScala

src/main/scala/dx/core/io/DxWorkerPaths.scala
jdidion: It is in dxCommon. "root" is the key used for caching.

src/main/scala/dx/core/io/DxWorkerPaths.scala
jdidion: They are used. It is where the job stdout, stderr, etc are stored.

commandlinegirl (commented):
The tests pass even though one workflow doesn't pass: I described it in this ticket: https://jira.internal.dnanexus.com/browse/APPS-266
It looks like it's unrelated to this change.

src/main/scala/dx/core/io/DxWorkerPaths.scala
jdidion: `import dx.util.{BaseEvalPaths, ExecPaths, Logger}`

jdidion: https://github.com/dnanexus/dxScala/blob/main/common/src/main/scala/dx/util/Paths.scala

src/main/scala/dx/core/ir/Type.scala
```diff
@@ -49,12 +50,7 @@ object Type {
   @tailrec
   def isPrimitive(t: Type): Boolean = {
     t match {
-      case TBoolean         => true
-      case TInt             => true
-      case TFloat           => true
-      case TString          => true
-      case TFile            => true
-      case TDirectory       => true
```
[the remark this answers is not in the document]
jdidion: TDirectory extends PrimitiveType, so it is primitive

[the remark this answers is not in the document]
commandlinegirl: Oh, right, I didn't notice `TDirectory` in line 20!

jdidion:
Hmm...I don't seem to be able to re-run the tests on this PR. Any idea why @commandlinegirl?

commandlinegirl:
I re-applied the `integration` label to kick off the tests and it worked. Checking what's up with unit tests

commandlinegirl:
BTW: this workflow `instance_types` is still failing (I'll check later why the test suite passes even though it fails, it doesn't look like a negative test):

image

jdidion:
I can look into that test. Looks like I should add a unit test there.

jdidion:
@commandlinegirl fixed now.

The reason that integration test job fails but the test passes is because the test is for dynamic selection of instance type. There is no way to access the actual instance type that was selected, other than to make an API call, which we currently don't have a way to do in the testing framework. We should add an issue to enable making assertions about attributes of the job/analysis for a test.

jdidion:
Unit tests are running and passing. The instance_types test is still failing, but for a different reason. I'll work on fixing that.

jdidion:
All integration tests passing 🎉
@commandlinegirl am I ok to merge?

commandlinegirl:
@jdidion Great, thanks for the instance types comaprison fix!

commandlinegirl:
@jdidion  Approved.

PR #11, opened 2020-11-11
APPS-275 Fix to compiling tools with syntax errors [WIP]

commandlinegirl:


jdidion:
Please don't make this change. We want to swallow the SyntaxException in order to let other factories have a chance to try to parse the document.

jdidion (commented):
Please don't make this change. We want to swallow the SyntaxException in order to let other factories have a chance to try to parse the document.

jdidion:
I will define a new exception type `NoSuchParserException` in wdlTools that will be thrown if a document is not detected to be parsable by any of the version-specific WDL parsers. Then we'll be able to distinguish between a document we do expect to be parsable but fails due to a syntax error vs one we do not expect to be parsable (e.g. a CWL document).

jdidion:
This is now fixed in #7

commandlinegirl:
@jdidion Thanks for the change! It'd be cleaner and easier to review if this was a separate PR rather than another part of #7.

PR #17, opened 2020-11-14
Migrate -streamFiles from dxWDL 1; update to new libraries

jdidion:
*  streamFiles
* update to wdlTools 0.10.6
* move API code to dxApi
* move DxFileAccessProtocol to separate protocols library
* fix formatting of executor templates (got messed up in a previous commit)
* move WDL type serialization code to wdlTools


commandlinegirl (changes requested):
@jdidion If possible, let's only add missing commits from dxWDL and leave further development (e.g. upgrading wdlTools) for the next release.

commandlinegirl:
@jdidion What is the reason for separating the code into more libraries? Please keep in mind code fragmentation into multiple packages makes a codebase more difficult to maintain, slower to develop, and read esp. for new developers. If it is not strictly necessary to split them I'd keep them as one bundle, also for the case of simplicity.

jdidion:
the code is used in more places than dxCompiler

commandlinegirl:
Where are they used? Where these packages were extracted from?

jdidion:
* dxCompiler has a dependency on wdlTools
* wdlTools is a stand-alone library/tool that can be used independently of dxCompiler, including in situations where DNAnexus is not used at all 
* both dxCompiler and wdlTools depend on common code in dxCommon
* dxCommon provides a plug-in system for file access protocols - currently we support DNAnexus and AWS, which are in the dxFileAccessProtocols package
* the DNAnexus protocol depends on the API, which is in the dxAPI package
* dxCompiler also depends on dxApi
* dxCommon, dxApi, and dxFileAccessProtocols are all in the dxScala repo
* So, at the repo level, wdlTools and dxCompiler both depend on dxScala

commandlinegirl:
1. What would be the scheme of reviewing, testing, and releasing of these repos?

E.g. a separate PR should be opened in dxScala to match PR #22 in dxCompiler. If a change is requested for that PR, a new version of the dxScala packages that #22 depends on will need to be created (and if other PRs depend on them, they’ll need to be rewritten and reissued)? Or first a PR is issued in dxScala, and once it’s approved, a pkg release is made and then a change is done in dxCompiler to bump up the dependencies versions and adjust the dxCompiler code (and if dxCompiler (and/or wdlTools) tests fail due to an error in dxScala, the cycle will be repeated)? Doe dxScala and wdlTools have tests?

2. Wouldn’t it be simpler to have one package corresponding to the dxScala repo instead of three, esp. since they also depend on each other?

3. A document/design doc would be great so we could comment on it. I created a table to better see the dependency structure here https://docs.google.com/document/d/1umR1PCNfQvjoMDeccjNsXvL0RMm2RJ1CP_ZzhK6TgBw/edit#heading=h.ctiubuc1ogo7, feel free to use/edit it.

jdidion:
1. Releasing these packages is quite simple - everything is handled by SBT with the exception of having to go to the sonatype server and click a couple of buttons. Whether the PR is submitted and reviewed in one repository vs another shouldn't matter.
2. The libraries do not all depend on each other - some can be (and are) used without the others. If I were going to merge any of them, I would put the dxFileAccessProtocols code in dxCommon - however, it has some heavy-weight dependencies (dxApi and the Java AWS library) that I don't want users of wdlTools to have to pull in if they don't need them. The purpose of modularity is to minimize the dependency graph, and thus the assembly size, for each of the tools.
3. I can work on this, but I don't have time for it now and I don't want to delay development. Let's make that a separate issue.

src/main/scala/dx/core/ir/Type.scala
```diff
@@ -25,6 +25,8 @@ object Type {
     */
   case class TOptional(t: Type) extends Type
 
+  sealed trait TCollection extends Type
```
commandlinegirl, line 28: Correct me if I'm wrong but TCollection doesn't seem to used anywhere so why is it introduced? Isn't it simpler for these types to be left Type?

commandlinegirl (changes requested):
Would it be possible to add a test for using streamFiles option?

build.sbt
```diff
@@ -79,12 +81,16 @@ val guavaVersion = "18.0"
...
+val commonsCompressVersion = "1.20"
 
 libraryDependencies ++= Seq(
+    "com.dnanexus" % "dxapi" % dxApiVersion,
     "com.dnanexus" % "dxcommon" % dxCommonVersion,
+    "com.dnanexus" % "dxfileaccessprotocols" % dxFileAccessProtocolsVersion,
     "com.dnanexus" % "wdltools" % wdlToolsVersion,
     "io.spray" %% "spray-json" % sprayVersion,
     "com.typesafe" % "config" % typesafeVersion,
+    "org.apache.commons" % "commons-compress" % commonsCompressVersion,
```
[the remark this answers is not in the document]
jdidion: that's a mistake - it's something that was introduced to support Archive format. I'll remove it.

src/main/scala/dx/core/ir/Type.scala
jdidion: I can remove it here. It's another thing that's needed for Archive format, but I will add it in that branch.

jdidion:
I can add a test for streamFiles at the IR level. It is actually very difficult to test for at the runtime level - I think it would require grepping the job log.

commandlinegirl (approved):

PR #21, opened 2020-11-17
create scatter job names using file names

jdidion:
Use the file name when creating names for scatter jobs.
This requires:
* dxApi 0.1.4, which caches the file name on DxFile if there is no describe
* dxFileAccessProtocols 0.1.2, which extracts the fileName from an "extended" dx URI and caches it on the DxFile


commandlinegirl:
What are `dxApi` and `dxFileAccessProtocols`?

commandlinegirl (commented):
Do we have a scatter test where we could add a check for job names?

jdidion:
Added a test

commandlinegirl (approved):
Thanks!

PR #22, opened 2020-11-17
upgrade to wdlTools 0.10.7

jdidion:
This fixes two bugs uncovered in RC5
- struct literals rendered incorrectly for v1
- array type-checking is too strict (can't coerce between empty and non-empty array types)


mhrvol (approved):
LGTM

PR #25, opened 2020-11-19
Apps-240 CWL -> IR 

mhrvol:


doc/CwlMapping.md
```diff
@@ -0,0 +1,37 @@
...
+* loadContents -> ?
+* loadListings -> ?
+
+* output.name -> name
+* output.types -> dxType
+* output.outputBinding -> ??
+
+## questions
+
+* WDL logger is in wdltools, how about cwl logger?
```
jdidion, line 34: The logger has moved to dxCommon

doc/CwlMapping.md
```diff
@@ -0,0 +1,37 @@
...
+* Extras are used to create IR application - I.e. InstanceType / container, correct? These are also given when starting the app
```
jdidion, line 35: No, extras are only given at compile time. Any information that is needed at runtime is added to the generated applet's `details` metadata.

src/main/scala/dx/core/ir/TypeSerde.scala
```diff
@@ -356,13 +356,11 @@ object TypeSerde {
       case TFile            => "File"
       case TDirectory       => "Directory"
       case THash            => "Hash"
+      case TEnum(values)    => s"Enum[${values.mkString(",")}]"
```
jdidion, line 359: I might prefer `Enum{X,Y,Z}` instead, as `{}` generally denotes a set

src/main/scala/dx/core/languages/Language.scala
```diff
@@ -1,11 +1,12 @@
 package dx.core.languages
 
+import org.w3id.cwl.cwl1_2.CWLVersion
 import wdlTools.syntax.WdlVersion
 import dx.util.Enum
 
 object Language extends Enum {
   type Language = Value
-  val WdlVDraft2, WdlV1_0, WdlV2_0, CwlV1_2 = Value
+  val WdlVDraft2, WdlV1_0, WdlV2_0, CwlV1_0, CwlV1_1, CwlV1_2 = Value
```
jdidion, line 9: This is not necessary - we will only support v1.2. Earlier versions must be upgraded via a pre-processing step. Going forward, we'll have to decide if we want to natively support additional versions, or only support the latest version and always require earlier versions to be upgraded.

src/main/scala/dx/core/languages/cwl/CwlUtils.scala
```diff
@@ -0,0 +1,154 @@
...
+        case e: CwlEnum => TSchema(e.name.get, Map[String, Type]("enum" -> TEnum(e.symbols)))
+        case _ => throw new Exception(s"Cannot convert CWL type ${cwlTypes.head} to IR")
+      }
+      case _ if cwlTypes.isEmpty => throw new IllegalArgumentException("Vector of CwlTypes must be non-empty!")
+      case other => throw new NotImplementedError(s"Multiple types are not supported yet! (${other} was given)")
+    }
+  }
+
+  def fromIRType(irType: Type): CwlType = {
+    def inner(innerType: Type): CwlType = {
```
jdidion, line 37: Use of an inner function isn't necessary here. I only use this pattern when the outer function has some additional arguments that I don't want to have to pass to the recursive call every time.

src/main/scala/dx/translator/cwl/CallableTranslator.scala
```diff
@@ -0,0 +1,72 @@
...
+                              locked: Boolean,
+                              defaultRuntimeAttrs: Map[String, Value],
+                              reorgAttrs: ReorgSettings,
+                              dxApi: DxApi = DxApi.get) {
+
+
+  def translateCallable(callable: Process, sourceFile: Path): Callable = {
+    callable match {
+      case tool: CommandLineTool =>
+        val taskTranslator = CwlToolTranslator.apply(tool, sourceFile)
```
jdidion, line 23: No need for the `.apply`. Calling `CwlToolTranslator(tool, sourceFile)` causes the `apply` function to be called.

src/main/scala/dx/translator/cwl/CallableTranslator.scala
```diff
@@ -0,0 +1,72 @@
...
+  def parseCommandInputParameter(input: CommandInputParameter): Parameter = {
+    val name = input.id.get.name.get
+    val dxType = CwlUtils.toIRType(input.types)
+    val defaultValue = Option(CwlUtils.toIRValue(input.default.getOrElse(CwlUtils.getDefaultCWLValue(input.types))))
+    val attributes: Vector[ParameterAttribute] = CwlUtils.createParameterAttributes(input.doc, input.label)
+
+    Parameter.apply(name.toString, dxType, defaultValue, attributes)
+  }
+
+  private case class CwlToolTranslator(tool: Process, sourceFile: Path) {
```
jdidion, line 47: `tool` should be a `CommandLineTool`

src/main/scala/dx/translator/cwl/CallableTranslator.scala
```diff
@@ -0,0 +1,72 @@
...
+
+    def apply(tool: CommandLineTool, sourceFile: Path): Application = {
```
jdidion, line 49: you don't need any arguments here - you already based them to the case class constructor.

src/main/scala/dx/translator/cwl/CallableTranslator.scala
```diff
@@ -0,0 +1,72 @@
...
+                              defaultRuntimeAttrs: Map[String, Value],
+                              reorgAttrs: ReorgSettings,
+                              dxApi: DxApi = DxApi.get) {
+
+
+  def translateCallable(callable: Process, sourceFile: Path): Callable = {
+    callable match {
+      case tool: CommandLineTool =>
+        val taskTranslator = CwlToolTranslator.apply(tool, sourceFile)
+        taskTranslator.apply(tool, sourceFile)
```
jdidion, line 24: you don't need the arguments to apply here - you already passed them to the constructor in the previous line

src/main/scala/dx/translator/cwl/CallableTranslator.scala
```diff
@@ -0,0 +1,72 @@
...
+      case _: Workflow => throw new NotImplementedError("Workflows are not implemented!")
+      case other => throw new NotImplementedError(s"Cwl type ${other.getClass} is not supported!")
```
jdidion, line 26: "CWL" should be all-caps in strings

jdidion (changes requested):
This is looking good. A few comments/changes. Due to the change of the project structure in #38 I will cherry-pick from this branch rather than merge the whole thing.

jdidion:
Thanks @mhrvol! All these changes have been incorporated into #40.

PR #28, opened 2020-11-20
update to wdlTools 0.11.0

jdidion:
This version fixes the issue type-checking struct-typed objects that Eugene discovered in RC5.

It also adds hooks for user-defined functions, which is needed for the new archive/unarchive operations we're adding to work around issues with very large numbers of outputs from deeply nested scatters.

https://github.com/dnanexus-rnd/wdlTools/pull/124


commandlinegirl (approved):
LGTM, thank you!

PR #38, opened 2020-12-02
Reorganize dxCompiler code into subprojects

jdidion:
* Splits codebase into 5 subprojects:
  - core: shared code between front-end and back-end
  - compiler: the front-end
  - executorCommon: shared code between executors
  - executorWdl: the WDL executor
  - executorCwl: the CWL executor
* A few classes were moved into different packages to clear up invalid dependencies (e.g. code in `core` depending on a class in `compiler`)
* Changes (almost) all references from dxWDL to dxCompiler
* Updates build and test scripts to handle the new layout
  - Changes the way the applet_resources folder is used:
    - Common binaries are stored in sub-folders by version
    - Each executor is built in its own sub-folder
* Updates to latest wdlTools (0.11.8)
* Note:
  - many test resources are duplicated because I can't figure out how to get sub-projects to share test resources
  - currently, a single version is used for all subprojects, located in core/src/main/resources/application.conf. All other subprojects have their own application.confs, but I'm not sure whether it's necessary to maintain the version information there; I have left it as-is for now.


compiler/src/main/resources/dxCompiler_runtime.conf
```diff
@@ -0,0 +1,8 @@
+dxCompiler {
+  regionToProject = [
+    {
+      region = "aws:us-east-1"
+      path = "dxCompiler_playground:/builds/jdidion/2.0.0-rc6"
```
commandlinegirl, line 5: Was this added accidentally?

commandlinegirl (commented):
Partially reviewed, just one question so far about `dxCompiler_runtime.conf`.

compiler/src/main/resources/dxCompiler_runtime.conf
jdidion: yes thanks for catching. I've updated .gitignore.

commandlinegirl (approved):

commandlinegirl:
@jdidion Please include the JIRA ticket ID in the commit title. Thanks!

commandlinegirl:
One more question: will anything need to change in the release process?

build.sbt
```diff
@@ -1,43 +1,190 @@
...
+    ),
+    publishMavenStyle := true,
+    // Tests
+    // If an exception is thrown during tests, show the full
+    // stack trace, by adding the "-oF" option to the list.
+    Test / testOptions += Tests.Argument("-oF"),
+    Test / parallelExecution := false
+    // Coverage
+    //
+    // sbt clean coverage test
```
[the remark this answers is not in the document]
jdidion: The goal is to re-enable it at some poInt

jdidion:
Not really any change to the release - there will be separate dxCompiler.jar and dxExecutor*.jar files, but it is only necessary to release the dxCompiler.jar on github.

PR #39, opened 2020-12-02
APPS-291 unit-tests GHA fix

mhrvol:
Problem with paralel tests was probably caused due to two sets of tests creating/deleting applets folder which tests use. Currently no applets are removed and folders are named using timestamp and UUID (timestamp for easier debugging when browsing filesystem and UUID is used in case two tests were started at the same time)

Applets-folder-deletion mechanism is changed from "Deleted when starting" to "deleted after tests are finished. Is this fine?

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -1,5 +1,6 @@
 name: Unit Tests
 
+
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


commandlinegirl (approved):
LGTM, thanks for the fix!

PR #44, opened 2021-01-12
APPS-324 Add release action

jdidion:
Adds a GHA for releases. Right now the only difference is that it runs the large integration test suite, as opposed to "integration" which runs the medium test suite. In the future, this action could be improved to automatically create GitHub releases for artifacts.


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+    types:
+      - opened
+      - synchronize
+      - ready_for_review
+      - labeled
+
+jobs:
+  test:
+    name: dxCompiler Integration Tests (Staging)
+    if: contains(github.event.pull_request.labels.*.name, 'integration')
```
commandlinegirl, line 19: We don't want to run the release and large tests with every commit to a PR labelled 'integration' . So how about for the release a label 'release' needs to be used?

Perhaps there is even a better trigger available - one where we don't need to create a PR to run a release.

.github/workflows/integration.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   test:
-    name: WDL Integration Tests (Staging)
+    name: dxCompiler Release Tests (Staging)
```
commandlinegirl, line 18: This should rather be dxCompiler Integration Tests (Staging) and github/workflows/release.yml  -> dxCompiler Release(Staging)

commandlinegirl (changes requested):
Thanks for this starter task! We'll need to take into account issues mentioned in  https://jira.internal.dnanexus.com/browse/APPS-324 when following up on this.

jdidion:
The test names have changed. After "dxCompiler Unit Tests" and " dxCompiler Integration Tests" pass, we'll need to force the merge and then update branch protection rules to disable "Scala Unit Tests" and "WDL Integration Tests".

commandlinegirl:
> The test names have changed. After "dxCompiler Unit Tests" and " dxCompiler Integration Tests" pass, we'll need to force the merge and then update branch protection rules to disable "Scala Unit Tests" and "WDL Integration Tests".

After updating the `main` branch protection rules already, the PR only requires the newly named pipelines to pass so I don't think we need to force merge:

image

commandlinegirl:
I marked the old ones back so other current PRs can pass.

PR #48, opened 2021-01-13
Null array as empty

jdidion:
DNAnexus does not distinguish between null and empty fo array inputs. Currently, if an empty array is passed to a non-optional WDL array input there is an error, even if the WDL array is allowed to be empty. This change (which depends on a new version of wdlTools) substitutes an empty array for a null value in the case of a non-optional array that is allowed to be empty.


commandlinegirl (approved):
LGTM, thanks!

r-i-v-a (approved):

r-i-v-a (changes requested):
Test failure message 

Error:  /home/runner/work/dxCompiler/dxCompiler/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala:297:84: unknown parameter name: strict
Error:          val callInputs = callIO.inputsFromValues(inputWdlValues, evaluator, strict = true)
Error:                                                                                     ^
Error:  one error found

commandlinegirl (changes requested):
The `inputsFromValues` here also need to be updated: https://github.com/dnanexus/dxCompiler/blob/a173927bb3b1f033eef6501b6c743e64acf57723/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala#L297

jdidion:
@commandlinegirl @r-i-v-a fixed - please re-review

commandlinegirl (approved):

PR #56, opened 2021-01-29
Fix APPS-369 and APPS-370

jdidion:
Fixes two bugs:
1. APPS-369 - structs out of order in generated code. This bug was due to an unnecessary conversion of a Vector to Map in CodeGenerator, which disordered the items.
2. APPS-370 - unable to resolve nested field references. There were two issues here:
    * Unnecessarily including outputs that are direct pass-throughs of inputs in the closure when determining the inputs to an Output stage.
    * A bug in the resolution of deeply nested field references in wdlTools.eval.Eval that is fixed in 0.11.18.


commandlinegirl (approved):
LGTM

PR #59, opened 2021-02-08
APPS 384 - fix expression evaluation errors

jdidion:
Fixes two errors uncovered by test case in APPS-384:
1. expression evaluator was not handling values wrapped in V_Optional
2. compound references were not handled correctly - a reference like `row.left` in an expression would only be evaluated if the name "row.left" were in the evaluation context. Now, if it is missing (say "row" is in the evaluation context, but not "row.left"), `row.left` is turned into an expression and evaluated using the expression evaluator.

Updates to wdlTools 0.12.3. This update also enforces uniqueness of variable names within the same scope.


executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala
```diff
@@ -46,7 +46,7 @@ object WdlWorkflowExecutor {
     val tasks = doc.elements.collect {
       case task: TAT.Task => task.name -> task
     }.toMap
-    WdlWorkflowExecutor(workflow, versionSupport, tasks, typeAliases.toMap, jobMeta)
+    WdlWorkflowExecutor(doc.source, workflow, versionSupport, tasks, typeAliases.toMap, jobMeta)
```
commandlinegirl, line 49: Would it be possible to add a short docstring, explaining esp what doc & docSource are?

commandlinegirl (approved):

executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala
jdidion: done

commandlinegirl (approved):

PR #60, opened 2021-02-10
APPS-389: handle optional non-file object inputs in native app(let)s

jdidion:
Currently, dxni skips any app(let)s with unsupported parameter types, but in the case of optional unsupported types, we can just ignore them.

This also updates to dxApi 0.1.9 to fix a bug with app (and globalworkflow) object ID parsing.

### DIFF
--- a/compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
+++ b/compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
@@ -77,7 +77,19 @@ case class WdlNativeInterfaceGenerator(wdlVersion: WdlVersion,
   private def wdlTypeFromDxClass(appletName: String,
                                  argName: String,
                                  ioClass: DxIOClass.Value,
-                                 isOptional: Boolean): WdlTypes.T = {
+                                 isOptional: Boolean): Option[WdlTypes.T] = {
+    if (ioClass == DxIOClass.Other) {
+      if (isOptional) {
+        logger.warning(s"ignoring applet ${appletName} optional non-file object input ${argName}")
+        return None
+      } else {
+        throw new Exception(
+            s"""|Cannot call applet ${appletName} from WDL, argument ${argName}
+                |has non-optional non-file object input ${argName}""".stripMargin
+              .replaceAll("\n", " ")
+        )
+      }
+    }
     val (t, isArray) = ioClass match {
       case DxIOClass.Boolean      => (WdlTypes.T_Boolean, false)
       case DxIOClass.Int          => (WdlTypes.T_Int, false)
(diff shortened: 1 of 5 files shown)


compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
```diff
@@ -77,7 +77,19 @@ case class WdlNativeInterfaceGenerator(wdlVersion: WdlVersion,
...
-                                 isOptional: Boolean): WdlTypes.T = {
+                                 isOptional: Boolean): Option[WdlTypes.T] = {
+    if (ioClass == DxIOClass.Other) {
+      if (isOptional) {
+        logger.warning(s"ignoring applet ${appletName} optional non-file object input ${argName}")
+        return None
+      } else {
+        throw new Exception(
+            s"""|Cannot call applet ${appletName} from WDL, argument ${argName}
+                |has non-optional non-file object input ${argName}""".stripMargin
```
commandlinegirl, line 88: Perhaps easier to read would be: "has required non-file object"

commandlinegirl: This can be understood as primitive required types are also not allowed. Since we don't recognize the type, can we at least say: "argument ${name} of this type.."?

compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
```diff
@@ -134,6 +144,7 @@ case class WdlNativeInterfaceGenerator(wdlVersion: WdlVersion,
       Some(createDnanexusStub(dxAppDesc.id, dxAppDesc.name, inputSpec, outputSpec))
     } catch {
       case e: Throwable =>
+        println(e)
```
commandlinegirl, line 147: This was left here by mistake?

commandlinegirl (commented):
Minor details

compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
jdidion: the type is "other" which is a catch-all for non-file object types. I put it there but I don't think it's very helpful.

compiler/src/main/scala/dx/dxni/wdl/WdlNativeInterfaceGenerator.scala
jdidion: yep good catch

commandlinegirl (approved):
LGTM

PR #62, opened 2021-02-11
APPS-344 Native tasks should use instance type from native app/let, not wrapper task.

r-i-v-a:
Don't use instance type from wrapper task on native task.
When generating task stub, include meta section with type=native for native task.


jdidion (approved):
LGTM

PR #66, opened 2021-02-17
APPS-395 - optionally use separate output folder per call

jdidion:
Currently, all output files for an analysis are placed in the same output folder. This can lead to name collisions, e.g. when a scatter job produces a file with a generic name, and then a downstream task tries to localize an array of those files (which all have the same name). This PR adds an option (`-separateOutputs`) to put the output files from each job in a separate folder.

Also updates to dxApi 0.1.10.


executorCommon/src/main/scala/dx/executor/WorkflowExecutor.scala
```diff
@@ -108,16 +108,34 @@ abstract class WorkflowExecutor[B <: Block[B]](jobMeta: JobMeta) {
...
     */
   protected def launchJob(executableLink: ExecutableLink,
                           name: String,
                           inputs: Map[String, (Type, Value)],
                           nameDetail: Option[String] = None,
-                          instanceType: Option[String] = None): (DxExecution, String) = {
+                          instanceType: Option[String] = None,
+                          folder: Option[String] = None): (DxExecution, String) = {
     val jobName: String = nameDetail.map(hint => s"${name} ${hint}").getOrElse(name)
+    val outputFolder = if (separateOutputs) {
```
commandlinegirl, line 127: Cannot jobName be non-unique for a set of calls within an analysis?

commandlinegirl (approved):
This LGTM, but I will look again at it Wed morning.

executorCommon/src/main/scala/dx/executor/WorkflowExecutor.scala
jdidion: It will always be unique within a given scope, and folders are nested by scope, so there should never be a folder name collision.

commandlinegirl (approved):

PR #68, opened 2021-02-19
APPS-399 - update to wdlTools 0.12.5

jdidion:
Currently, given a command such as

```
echo ~{a} \
  ~{b}
echo hello
```

If `b` is optional and unspecified, this will result in the command:

```
echo foo \
echo hello
```

Because blank lines are (incorrectly) removed. This bug is fixed in wdlTools 0.12.5, so we just update it in dxCompiler.


commandlinegirl (approved):
LGTM

PR #70, opened 2021-02-23
Ignore default values in native app stubs

jdidion:
WDL requires that all output declarations have a value assigned, but in effect these values are ignored for native apps. This PR changes the WDL translator to not eval add default values to the IR `Parameter` instances for native app stub output parameters, because these can cause issues when they are copied.


commandlinegirl (approved):
LGTM

PR #72, opened 2021-02-24
APPS-324 Simplify release pipeline; add docs

commandlinegirl:
* Builds a release directly from main, doesn't push any branch back
* Runs large integration tests by default
* Adds input validation
* Removes release entry rollback (so that the drafted page is kept in case of failure)


jdidion (approved):
LGTM

PR #74, opened 2021-02-24
use wfName instead of wf.name in call to createCommonApplet

jdidion:
Using the wrong name causes applet name collisions, which causes compilation to fail for complex nested workflows.


commandlinegirl (approved):
LGTM but is it the only place where `wf.name` shouldn't be used in tis file?

PR #75, opened 2021-03-02
APPS-419 handle optional call parameters with null arguments

jdidion:
Currently a workflow like the following fails with an error. This PR fixes the handling of null values specified for optional call parameters.

```wdl
version 1.1

workflow Test {
  input {
    Int? opt
  }

  call Foo {
    input: opt=opt
  }
}

task Foo {
  input {
    Int? opt = 20
  }
  command <<<
  echo ~{default="20" opt}
  >>>
}
```


commandlinegirl (approved):
LGTM, thanks for the fix!

PR #79, opened 2021-03-10
APPS-288 - fixes issue with using both streaming and non-streaming files

jdidion:
The issue was in the generated job script in the bash code block for dxda - it cd's to / at the beginning and then is supposed to cd back to the home directory at the end, but instead it was cd-ing to the work dir. This only causes an issue if dxfuse is also used, because it tries to source environment, which is in the home dir. The error was not showing up in the log because we pipe the output from `source environment` to /dev/null to avoid leaking the token to the logs, so it was also swallowing the error message.

Also fixes the parent directory name written to the dxfuse manifest (removes the extra leading '/').

Also updates to dxCommon 0.2.10 which fixes another issue - the LocalizationDisambiguator was creating directories in the dxfuse mount point, which is wrong behavior, but I guess dxfuse doesn't complain as long as the directory is empty.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 ## in dev
 
 * Adds `-useManifests` option to generate applets and workflows whose inputs and outputs are manifest files
+* Fixes issue with using both streaming and non-streaming file inputs in the same task
 
 ## 2.3.1 03-03-2021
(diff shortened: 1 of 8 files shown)


commandlinegirl (approved):
LGTM, thanks for the fix!

PR #81, opened 2021-03-15
APPS-456 fixes issue with having a scatter as the first workflow element

jdidion:
Computed inputs (i.e. scatter variables) need to be excluded from workflow inputs.


commandlinegirl (approved):
LGTM

PR #83, opened 2021-03-17
don't log docker passwords

jdidion:
this change avoids echoing docker passwords to the job log


commandlinegirl (approved):
This LGTM but it can be worth it to disable it in the generic preamble https://github.com/dnanexus/dxCompiler/blob/main/compiler/src/main/resources/templates/generic_docker_preamble.ssp#L25

jdidion:
It is disabled in the generic preamble

PR #84, opened 2021-03-18
Update to wdlTools 0.12.7

jdidion:
This version relaxes a few constraints to enable parsing of some Broad workflows that use non-compliant WDL syntax.


commandlinegirl (approved):
LGTM

PR #88, opened 2021-03-25
use static AWS client, don't echo password

jdidion:
There are still intermittent issues with installing the AWS CLI at runtime. This PR packages the AWS CLI installer with the executor. Also fixes echoing of passwords to the log (for real this time).

### DIFF
--- a/compiler/src/main/resources/templates/generic_docker_preamble.ssp
+++ b/compiler/src/main/resources/templates/generic_docker_preamble.ssp
@@ -19,17 +19,17 @@ if [[ ${bashDollar}num_lines != 1 ]]; then
     exit 1
 fi
 dx download ${bashDollar}{DOCKER_CREDENTIALS} -o ${bashDollar}{HOME}/docker_credentials
+
+# disable echo to avoid logging password
+set +e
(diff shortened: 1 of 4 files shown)


compiler/src/main/resources/templates/generic_docker_preamble.ssp
```diff
@@ -19,17 +19,17 @@ if [[ ${bashDollar}num_lines != 1 ]]; then
     exit 1
 fi
 dx download ${bashDollar}{DOCKER_CREDENTIALS} -o ${bashDollar}{HOME}/docker_credentials
+
+# disable echo to avoid logging password
+set +e
```
commandlinegirl, line 24: Shouldn't it be set +x / set -x for printing out commands, instead of +e/-e which exits upon errors.

jdidion: oops - you're right, good catch

commandlinegirl: Np, sorry I missed it the first time.

commandlinegirl (approved):
LGTM, thank you for the quick fix.

PR #94, opened 2021-04-16
APPS-534 poll analysis on reorg

jdidion:
The analysis is updated in the backend asynchronously, so we need to check the `dependsOn` field to determine if it is still waiting on the output of a previous stage before we can proceed with the reorg.

This PR also updates to the latest dxScala dependencies.

### DIFF
--- a/doc/CustomReorgAppletExample.md
+++ b/doc/CustomReorgAppletExample.md
@@ -77,7 +77,7 @@ def main(reorg_conf___=None, reorg_status___=None):
     while True:
         analysis_desc = dxpy.describe(analysis_id)
         if analysis_desc.get("dependsOn"):
-            time.sleep(1)
+            time.sleep(3)
         else:
             break
             
--- a/doc/CustomReorgAppletExample.md
+++ b/doc/CustomReorgAppletExample.md
@@ -59,7 +59,7 @@ The applet code does the following.
 
 ```
 import json
-
+import time
 import dxpy
 
 @dxpy.entry_point('main')
@@ -73,7 +73,15 @@ def main(reorg_conf___=None, reorg_status___=None):
 
     # find the output stage of the current analysis
     analysis_id = dxpy.describe(dxpy.JOB_ID)["analysis"]
-    stages = dxpy.describe(analysis_id)["stages"]
+    # describe the analysis in a loop until dependsOn is empty
+    while True:
+        analysis_desc = dxpy.describe(analysis_id)
+        if analysis_desc.get("dependsOn"):
+            time.sleep(1)
+        else:
+            break
+            
+    stages = analysis_desc["stages"]
 
     # retrieve the dictionary containing outputs, where key is the name of output and value is the link to the file.
     output_map = [x['execution']['output'] for x in stages if x['id'] == 'stage-outputs'][0]
--- a/doc/CustomReorgAppletExample.md
+++ b/doc/CustomReorgAppletExample.md
@@ -59,7 +59,6 @@ The applet code does the following.
 
 ```
 import json
-import time
 import dxpy
 
 @dxpy.entry_point('main')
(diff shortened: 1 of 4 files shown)


doc/CustomReorgAppletExample.md
```diff
@@ -73,7 +73,15 @@ def main(reorg_conf___=None, reorg_status___=None):
 
     # find the output stage of the current analysis
     analysis_id = dxpy.describe(dxpy.JOB_ID)["analysis"]
-    stages = dxpy.describe(analysis_id)["stages"]
+    # describe the analysis in a loop until dependsOn is empty
+    while True:
+        analysis_desc = dxpy.describe(analysis_id)
+        if analysis_desc.get("dependsOn"):
+            time.sleep(1)
```
commandlinegirl, line 80: Is one second enough to wait? It usually takes more than 1s to describe. Won't the user eventually get throttled?

doc/ExpertOptions.md
```diff
@@ -877,6 +877,8 @@ it may misplace or outright delete files. The applet:
 3. has to be careful about inputs that are *also* outputs. Normally, these should not be moved.
 4. should use bulk object operations, so as not to overload the API server.
 
+You must also be aware that the analysis information is updated in the platform's database asynchronously, so the result of calling `dx describe` on the analysis may not be up-to-date. The most reliable method for making sure you have an up-to-date analysis description is to call `dx describe` in a loop (waiting at least 1 second between iterations), and exit the loop when the `dependsOn` field of the analysis is an empty array.
```
commandlinegirl, line 880: suggestion: 1 second -> a few seconds

executorCommon/src/main/scala/dx/executor/WorkflowExecutor.scala
```diff
@@ -198,7 +199,20 @@ abstract class WorkflowExecutor[B <: Block[B]](jobMeta: JobMeta, separateOutputs
...
+    // the analysis is updated asynchronously, so we need to check the dependsOn field
+    // to determine if it is still waiting on the output of a previous stage before we
+    // can proceed with the reorg
+    val desc = Iterator
+      .continually(
+          analysis.describeNoCache(Set(Field.Input, Field.Output, Field.DependsOn))
+      )
+      .collectFirstDefined {
+        case a if a.dependsOn.exists(_.nonEmpty) =>
+          Thread.sleep(1000)
```
commandlinegirl, line 211: s/1000/3000

commandlinegirl (changes requested):

commandlinegirl (approved):
LGTM, thank you!

r-i-v-a (approved):

executorCommon/src/main/scala/dx/executor/WorkflowExecutor.scala
```diff
@@ -198,7 +199,23 @@ abstract class WorkflowExecutor[B <: Block[B]](jobMeta: JobMeta, separateOutputs
...
+    val desc = Iterator
+      .continually(
+          analysis.describeNoCache(Set(Field.Input, Field.Output, Field.DependsOn))
+      )
+      .collectFirstDefined { a =>
+        a.dependsOn match {
+          case Some(Vector(jobId)) if jobId == jobMeta.jobId => Some(a)
+          case None | Some(Vector())                         => Some(a)
+          case _ =>
+            Thread.sleep(3000)
```
mhrvol, line 214: Cant we get stuck in infinite loop in case jobId wouldn't change for some reason? Or is it impossible not to pass the jobId == jobMeta.jobId condition in the end?

I was wondering whether we want to have some kind of timeout here - just to be safe

jdidion: Yes it's possible. I asked this question in GChat and Stanley commented that the job will eventually terminate due to timeout, so it's ok to retry without a limit. I still think a max number of retries is a good idea. I'll bring it up at stand-up and see what Eugene thinks.

jdidion:
The analysis is updated in the backend asynchronously, so we need to check the dependsOn field to determine if it is still waiting on the output of a previous stage before we can proceed with the reorg.

This PR also updates to the latest dxScala dependencies.

jdidion:
The analysis is updated in the backend asynchronously, so we need to check the dependsOn field to determine if it is still waiting on the output of a previous stage before we can proceed with the reorg.

This PR also updates to the latest dxScala dependencies.

PR #95, opened 2021-04-19
APPS-512 - add private variables from outer scope to env when computing inputs of nested scatters

jdidion:
The problem is illustrated by this WDL snippet:

```
    Array[Int] ints1 = [1, 2, 3]
    Array[Int] ints2 = [10, 11]

    scatter (x in ints1) {
        scatter (y in ints2) {
            call xxx_add{ input: a = x, b = y }

            Int i = ints1[x-1]
            if (i < 3) {
                String s = "hello ${x} ${y}"
            }
        }
    }
```

Currently, `ints1` and `ints2` are not added to the environment when translating the inner scatter, so the generated applet is missing inputs for those. This PR fixes the issue.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## in develop
+
+* Fixes issue with nested scatters that reference private variables from outer scopes
+
 ## 2.4.1 03-25-2021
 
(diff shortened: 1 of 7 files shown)


commandlinegirl (approved):
LGTM, thanks for the regression tests and the comments in the code!

PR #102, opened 2021-04-21
APPS-547 Add update_version script that works on macOS

r-i-v-a:
Noticed that the existing update_version.sh doesn't work on macOS because of a difference in macOS sed [https://stackoverflow.com/questions/19456518/invalid-command-code-despite-escaping-periods-using-sed/19457213#19457213]. Added a script adapted for macOS, left the original one in place.

Can be used e.g. for bumping the working version on `develop` to the next `-SNAPSHOT` version.


jdidion:
A separate script is unnecessary. The command `sed -i'' -e "s/version.*$/version = \"${VERSION}\"/" $i` will work on both mac and linux.

scripts/update_version.sh
```diff
@@ -8,6 +10,9 @@ CONF_FILES=(
...
     ./executorCommon/src/main/resources/application.conf
 )
+
+# Adapted to work also w/ macOS version of sed, where -i expects extension arg
+# https://stackoverflow.com/questions/19456518/invalid-command-code-despite-escaping-periods-using-sed/19457213#19457213
 for i in ${CONF_FILES[@]}; do
-    sed -i "s/version.*$/version = \"${VERSION}\"/" $i
-done
\ No newline at end of file
+    sed -i'' -e "s/version.*$/version = \"${VERSION}\"/" $i
```
r-i-v-a, line 17: @jdidion @commandlinegirl   Updated, please review

jdidion (approved):

r-i-v-a:
@jdidion @commandlinegirl   I think I need to make the WDL integration tests run on this PR

commandlinegirl:
No need to run integration tests since this change didn't affect the compiler code and this script is not used in the tests anywhere. You can go ahead and merge. 

If you remove the `integration` label, you should be able to merge again.

PR #104, opened 2021-04-22
APPS 516 - error dereferencing struct as a call argument

jdidion:
In dxCompiler, a "trivial" expression is one that does not require evaluation to determine the value, i.e. it is a primitive value, a collection of primitives, or a reference to an input variable. We had also considered access to a field of a struct or a call return value as trivial, but this will not be the case for applets that are called directly as a stage in a workflow.

For example, in the following workflow, the applet for the `test` task takes a String input, but the call argument has to be evaluated at runtime, so we have to pass the struct (rather than the field value). Due to the mismatch between the call argument and the task input, a wrapper task is needed to evaluate the expression at runtime, which makes the expression non-trivial.

```
version 1.0

struct SampleStruct {
  String sample_name
}

workflow struct_deref {
  input {
    SampleStruct sampleStruct
  }

  call test {
    input:
      sample_name = sampleStruct.sample_name
  }

  output {
    String out = test.out
  }
}

task test {
  input {
    String sample_name
  }

  command <<<
    echo ~{sample_name}
  >>>

  output {
    String out = read_string(stdout())
  }
}
```


core/src/main/scala/dx/core/languages/wdl/WdlUtils.scala
```diff
@@ -911,6 +914,10 @@ object WdlUtils {
         innerElements: Vector[TAT.WorkflowElement]
     ): Vector[TAT.OutputParameter] = {
       innerElements.flatMap {
+        // TODO: in the case of private variables where the expression access a
```
commandlinegirl, line 917: s/access/accesses

commandlinegirl (approved):

PR #113, opened 2021-05-06
APPS-579 Add regression tests for expression evaluation: boolean, substitution.

r-i-v-a:


scripts/clean_build.sh
```diff
@@ -11,9 +11,7 @@ rm dx*.jar
 # Clean artifacts from prev. builds, platform
 username=`dx whoami --id`
 dx rm -r dxCompiler_playground:/builds/$username
-dx rm -r dxCompiler_playground:/builds/user-$username
```
r-i-v-a, line 14: `dx whoami` with `--id` already returns the user-xxxx form of the username.

Looking at the implementation of dxApi.whoami(), it looks consistent with the dxpy one (both use the same system API call), so the unit_tests folder should also get the same name.

test/bugs/apps_579_boolean_flag_expr.wdl
```diff
@@ -0,0 +1,25 @@
+version 1.1
```
r-i-v-a, line 1: @jdidion   Should I actually make these 1.0 tests? Do we only test against 1.1 for new features not existing in 1.0?

r-i-v-a: From comment on other PR -- 1.1 is ok

r-i-v-a:
This one is ready for review @jdidion

jdidion (approved):

PR #122, opened 2021-05-20
Update to wdlTools 0.13.0

jdidion:
Also implements the ability to parse and type-check WDL in lenient mode, although the option is currently not available to the user.


jdidion:
It looks like there's another code generator bug - will have to fix this and make a new wdlTools release.

r-i-v-a (dismissed):

PR #130, opened 2021-06-02
APPS-658 dxCompiler will annotate the generated applets and workflows with their statically-determined dependencies and add them to details & description.

r-i-v-a:


compiler/src/main/scala/dx/compiler/ExecutableTree.scala
```diff
@@ -5,6 +5,7 @@ import dx.core.ir.{Application, Callable, ExecutableKind, ExecutableLink, Workfl
 import dx.util.CodecUtils
 import spray.json._
 
+// APPS-622 links contains callee applets
```
r-i-v-a, line 8: @jdidion   Actually, this is probably the callee generated applets in the tree of stuff being compiled -- will try some example w/ a native app dependency using dxni and look for that dependency, which should be annotated for APPS-622?

compiler/src/main/scala/dx/translator/CallableAttributes.scala
```diff
@@ -4,10 +4,12 @@ import dx.core.ir.{CallableAttribute, Value}
 
 object CallableAttributes {
   final case class TitleAttribute(text: String) extends CallableAttribute
+  // APPS-622 description attribute that will be shown in UI?
```
r-i-v-a, line 7: @jdidion   These are the fields that should be filled in for APPS-622?

compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala
```diff
@@ -765,14 +769,19 @@ case class CallableTranslator(wdlBundle: WdlBundle,
         }
       }
 
+      // APPS-622 Construct DetailsAttribute w/ dependency annotations,
```
r-i-v-a, line 772: @jdidion   Is this where to add a DetailsAttribute for a generated applet?

compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala
```diff
@@ -173,6 +173,9 @@ case class CallableTranslator(wdlBundle: WdlBundle,
       }
       val standAloneTask =
         WdlDocumentSource(codegen.createStandAloneTask(cleanedTask), versionSupport)
+
+      // APPS-622 Related to building applet for task,
```
r-i-v-a, line 177: @jdidion   Actually, is this where metadata goes into task-level applets?

r-i-v-a:
Added a test, and this is ready for review. @jdidion

compiler/src/main/scala/dx/translator/wdl/RuntimeTranslator.scala
```diff
@@ -77,7 +78,8 @@ case class RuntimeTranslator(wdlVersion: WdlVersion,
                              metaSection: Option[TAT.MetaSection],
                              defaultAttrs: Map[String, Value],
                              evaluator: Eval,
-                             dxApi: DxApi = DxApi.get) {
+                             dxApi: DxApi = DxApi.get,
+                             logger: Logger = Logger.get) {
```
jdidion, line 82: What's the purpose of adding `logger`?

compiler/src/main/scala/dx/compiler/ExecutableTree.scala
```diff
@@ -181,6 +181,41 @@ object ExecutableTree {
     }
   }
 
+  def processWorkflowNames(TreeJS: JsObject, namesAcc: Set[String]): Set[String] = {
```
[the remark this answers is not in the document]
jdidion: Should probably also use better names for the `processWorkflow` and `processApplets` functions above.

compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
```diff
@@ -297,6 +297,17 @@ class ExecutableCompiler(extras: Option[Extras],
     }
   }
 
+  // Merge initial description & extended description
+  private def descriptionToNative(description: Option[String],
+                                  extendedDescription: Option[String]): Map[String, JsValue] = {
+    (description, extendedDescription) match {
+      case (Some(d), Some(e)) => Map("description" -> JsString(d + "\n\n" + e))
```
jdidion, line 304: Use format strings rather than concatenation. `s"${d}\n\n${e}"`

jdidion: You can also do

```scala
Map("description" -> JsString(Vector(description, extendedDescription).flatten.mkString("\n\n")))
```

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
     (commonMeta ++ workflowMeta, commonDetails)
   }
 
+  private def listMd(s: String): String = {
+    "\n* " + s
```
jdidion, line 181: use format strings rather than concatenation

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+  }
+
+  private def listMd2(s: String): String = {
```
jdidion, line 184: Use a more descriptive name or at least put docstrings.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+    "\n    * `" + s + "`"
+  }
+
+  // Summarize reportable dependencies from details of a workflow
```
jdidion, line 188: Add a TODO that all this reporting code should be moved to templates.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+  private def reportableDependenciesFromWorkflow(
+      workflow: DxExecutable
```
jdidion, line 190: I think the parameter type should be `DxWorkflow` not `DxExecutable`.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+  ): Option[String] = {
+    val description = workflow.describe(Set(Field.Name, Field.Details, Field.Hidden)) match {
+      case w: DxWorkflowDescribe => w
+      case _                     => throw new Exception(s"Expected ${workflow.id} to be a workflow.")
+    }
+
+    // Do not generate description section if hidden
+    if (description.hidden.isDefined && description.hidden.get) {
```
jdidion, line 198: Replace with `description.exists(_)`

jdidion: Also this does nothing. You either need to `return` here or wrap the rest of the code in an `else` block.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+      None
+    }
+
+    // Generate Markdown description section
+    val headerMd = s"\n## Sub-workflow ${description.name}"
+    val details = description.details.collect {
```
jdidion, line 204: No need for two separate statements here. Just do `description.details match ...`

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+      case v: JsValue => v.asJsObject
+    }
+    val md = details match {
+      case Some(JsObject(details)) => {
+        val appsMd = details.get(Constants.NativeAppDependencies) match {
+          case Some(JsArray(a)) => listMd("Native apps") + a.map(v => listMd2(v.toString)).mkString
```
jdidion, line 210: use format string

jdidion: Acutally I'd just build up a nested Vector of everything you want to concatenate and then at the very end do a `.flatten.mkString`.

So

```scala
Vector(listMd("Native apps")) ++ a.map(v => listMd2(v.toString))
```

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+      case _ => ""
+    }
+    md match {
+      case s: String if !s.isEmpty => Some(headerMd + md)
+      case _                       => None
+    }
+  }
+
+  // Summarize reportable dependencies from details of an applet
+  private def reportableDependenciesFromApplication(
```
jdidion, line 234: same comments as above

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +177,168 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+    val topLevelMd = appsMd + filesMd + dockerRegistryCredentialsMd
+
+    val executableNames = ExecutableTree.getExecutableNames(execTree)
+    val executables = executableNames.collect { n =>
+      executableDict.get(n)
+    }.flatten
+
+    val subWorkflowsMd = executables
+      .collect {
+        case CompiledExecutable(Workflow(_, _, _, _, _, _, _, _, _, _, _), dxExec, _, _) => dxExec
```
jdidion, line 316: use `workflow: DxWorkflow` instead

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -414,7 +586,89 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+                      _
+                      ) =>
+                    dxExec.id
+                }
+              }
+              .filter(_.startsWith("app-"))
+        )
+        .toVector
+    val nativeAppDependenciesDetails = nativeAppDependencies match {
+      case nativeAppDependencies if nativeAppDependencies.length > 0 =>
```
jdidion, line 621: `nonEmpty`

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -414,7 +586,89 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+        Map(
+            Constants.NativeAppDependencies -> JsArray(
+                nativeAppDependencies.map(id => JsString(id))
+            )
+        )
+      case _ => Map.empty
+    }
+
+    // Collect file dependencies from inputs & private variables
+    val fileInputParams = workflow.inputs.map(i => i._1).filter(param => param.dxType == Type.TFile)
```
jdidion, line 631: You're going to miss optional types. Do `Type.unwrapOptional(param.dxType)`.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -414,7 +586,89 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+    val fileDependencies =
+      Set
+        .from(
+            fileInputParams
```
jdidion, line 635: You can do `(vecA ++ vecB).distinct`

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -414,7 +586,89 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+              .flatMap(CompilerUtils.fileDependenciesFromParam) ++ workflow.varFileDependencies
+        )
+        .toVector
+    val fileDependenciesDetails = fileDependencies match {
```
jdidion, line 639: The match here isn't necessary. Easier/clearer to do

```scala
if (fileDependencies.nonEmpty) { ... } else { Map.empty }
```

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -414,7 +586,89 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+        Map(Constants.FileDependencies -> JsArray(fileDependencies.map(s => JsString(s))))
+      case _ => Map.empty
+    }
+
+    // Collect Docker registry credentials URI
+    val dockerRegistryCredentialsUri: Option[String] = extras.flatMap(_.dockerRegistry) match {
+      case Some(DockerRegistry(_, credentials, _, _)) => Some(credentials)
+      case _                                          => None
+    }
+    val dockerRegistryCredentialsUriDetails = dockerRegistryCredentialsUri match {
```
jdidion, line 650: Here you can do `dockerRegistryCredentialsUri.map(cred => Map(Constants.DockerRegistryCredentialsUri -> JsString(cred))).getOrElse(Map.empty)`

compiler/src/main/scala/dx/compiler/package.scala
```diff
@@ -0,0 +1,37 @@
+package dx.compiler
+
+import dx.core.ir.{Parameter, Value}
+import dx.translator.ParameterAttributes
+
+package object CompilerUtils {
+
+  // Return a parameter's dependencies on non-local files
+  def fileDependenciesFromParam(param: Parameter): Set[String] = {
```
jdidion, line 9: I would just put this as a protected method in the parent class `ExecutableCompiler`.

core/src/main/scala/dx/core/Constants.scala
```diff
@@ -3,22 +3,28 @@ package dx.core
...
-  val ExecLinkInfo: String = "execLinkInfo"
-  val BlockPath: String = "blockPath"
-  val WfFragmentInputTypes: String = "fqnDictTypes"
-  val InstanceTypeDb: String = "instanceTypeDB"
+  // keys used in details of native applets, workflows
+  val ExecLinkInfo = "execLinkInfo"
+  val BlockPath = "blockPath"
+  val WfFragmentInputTypes = "fqnDictTypes"
+  val InstanceTypeDb = "instanceTypeDB"
+  val DxInstanceType = "dxInstanceType"
```
jdidion, line 11: I'd call this `staticInstanceType` to make it clear how it's different than just getting `instanceType` from the applet's runtimeExec.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -248,6 +259,46 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
     (commonMeta ++ applicationMeta, commonDetails ++ taskSpecificDetails)
   }
 
+  private def listMd(s: String): String = {
```
jdidion, line 262: These are redundant with WorkflowCompiler - you can make the protected methods in the shared parent class `ExectuableCompiler`.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -248,6 +259,46 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
...
+    "\n* " + s
+  }
+
+  private def listMd2(s: String): String = {
+    "\n    * `" + s + "`"
+  }
+
+  // Summarize dependencies that are not bundled for cloning
+  // with the app/let for adding to description in Markdown
+  private def summarizeReportableDependencies(
```
jdidion, line 272: Same comments as in WorkflowCompiler

jdidion (changes requested):

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
r-i-v-a: Took your 2nd suggestion.

compiler/src/main/scala/dx/compiler/ExecutableTree.scala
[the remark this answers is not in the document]
r-i-v-a: Made methods private, better names for my & existing methods.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Added TODO comment

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Also: same for `DxApplet` instead of `DxExecutable`

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +176,161 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+    val executableNames = ExecutableTree.extractExecutableNames(execTree)
+    val executables = executableNames.collect { n =>
+      executableDict.get(n)
+    }.flatten
+
+    val subWorkflowsMd = executables
+      .collect {
+        case CompiledExecutable(Workflow(_, _, _, _, _, _, _, _, _, _, _), dxExec, _, _) =>
+          dxExec match {
+            case w: DxWorkflow => w
```
r-i-v-a, line 305: Please have a look at DxWorkflow, DxApplet here. Is this what you meant, or is there a better way? @jdidion

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Done (in parent class).

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Can you check the code for this part -- is that what you meant, or is there a better way?

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Like this? 

```
    if (description.hidden.exists(h => h)) {
      return None
    }
```

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Somehow, when I try 

```
    val details = description.details match {
      case Some(JsValue(v)) => v.asJsObject match {
        ...
      }
    }
```

I get a compilation error that JsValue can't be found. This is despite the import statement still being 

```
import spray.json._
```

Any idea why that happens?

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Doing flatten.mkString at the end of that would change the logic because you are always putting "Native apps," regardless of whether there is anything to go in the sub-list under it. My goal is to add the header for some type of dependency only when that type of dependency is present.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Good catch, done.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: you can put `dxExec: DxWorkflow` in the first case statement so you don't have to do nested matches.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: yes, although `exists(_)` is equivalent

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: `JsObject`, not `JsValue`

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: That would still be the case. Maybe you're not understanding the suggestion - might be easier to talk through in person.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: you can put `dxExec: DxWorkflow` in the first case statement so you don't have to do nested matches.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Thanks, that works.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: I think I'm misunderstanding the suggestion -- to discuss in person. @jdidion

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: Cool, I didn't realize `Option.map` method exists. I guess it makes sense if you think of Option as a special case of collection containing 0-1 items. Done. @jdidion

compiler/src/main/scala/dx/compiler/package.scala
r-i-v-a: Done.

compiler/src/main/scala/dx/translator/wdl/RuntimeTranslator.scala
r-i-v-a: Probably used it for testing and forgot to remove.

core/src/main/scala/dx/core/ir/RunSpec.scala
```diff
@@ -82,14 +82,17 @@ object RunSpec {
 
   /**
     * A task may specify a container image to run under. Currently, DNAnexus only
-    * supports Docker images. There are three supported options:
-    *   None:    no image
+    * supports Docker images. There are four supported options:
+    *   None: no image
+    *   Dynamic: Image determined dynamically at runtime, also assumed to reside
```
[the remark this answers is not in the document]
r-i-v-a: I think what I mean there is that it's possible it could require network access for download. Edited comment for clarity.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
jdidion: Another way you can do it:

```
Vector(
  Option.when(nativeAppDependencies.nonEmpty)(Vector(listMd("Native apps")) ++ a.map(v => listMd2(v.toString))),
  Option.when(...)
).flatten.flatten.mkString
```

r-i-v-a: I've used that style in WorkflowCompiler.summarizeReportableDependencies, but elsewhere (where also unwrapping JsValue types into Scala types) I think it's less readable not to use "match," so elsewhere I'm just going to edit for using interpolation instead of concatenation.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -167,6 +176,154 @@ case class WorkflowCompiler(extras: Option[Extras],
...
+            Vector(listMd("Files")) ++ fileDependencies.map(v => listMd2(v.toString))
+        ),
+        Option.when(dockerRegistryCredentialsUri.nonEmpty)(
+            Vector(listMd("Docker registry credentials file"),
+                   listMd2(dockerRegistryCredentialsUri.get))
+        )
+    ).flatten.flatten.mkString
+
+    val executableNames = ExecutableTree.extractExecutableNames(execTree)
+    val executables = executableNames.collect { n =>
```
jdidion, line 293: when the function you're mapping over takes a single argument, you can shorten it to `executableNames.collect(executableDict.get)`

compiler/src/main/scala/dx/translator/wdl/RuntimeTranslator.scala
```diff
@@ -4,6 +4,7 @@ import dx.api.{DxApi, DxPath, DxUtils, InstanceTypeRequest}
 import dx.core.ir.RunSpec._
 import dx.core.ir.{ExecutableKind, ExecutableKindNative, ExecutableType, RuntimeRequirement, Value}
 import dx.core.languages.wdl.{DxRuntimeHint, IrToWdlValueBindings, Runtime, WdlUtils}
+import scala.util.{Try}
```
jdidion, line 7: you don't need the `{}` if you're only importing one thing

jdidion (changes requested):
A few more minor comments. Otherwise looks good.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
r-i-v-a: ```executableNames.collect(executableDict.get(_)).flatten```

jdidion (approved):
Nice work!

PR #133, opened 2021-06-07
APPS-634 Fix example script run command, use same flags as in customer project.

r-i-v-a:


mhrvol (dismissed):
LGTM

btw why not brief anymore?

r-i-v-a:
> LGTM
> 
> btw why not brief anymore?

I can keep --brief, actually -- it will behave similarly but just print less. Was mostly bothered here by the last line ending in \ after the last edit. @mhrvol

jdidion (approved):

PR #145, opened 2021-07-01
APPS-677 Fix so that option -execTree pretty, json prints the workflow tree.

r-i-v-a:


jdidion (changes requested):
The compile command also needs to print the compiled object IDs. So you need to override `message` on `SuccessfulCompileNativeWithJsonTree` and `SuccessfulCompileNativeWithPrettyTree` and do something like

```
override def message: String = {
    s"${executableIds.mkString(",")}\n${prettyTree}"
}
```

r-i-v-a:
> The compile command also needs to print the compiled object IDs. So you need to override `message` on `SuccessfulCompileNativeWithJsonTree` and `SuccessfulCompileNativeWithPrettyTree` and do something like
> 
> ```
> override def message: String = {
>     s"${executableIds.mkString(",")}\n${prettyTree}"
> }
> ```

Oh right. I'm used to seeing the IDs printed last, so I'll put them there.

jdidion (approved):

PR #161, opened 2021-08-05
return error on empty return code file

emiloslavsky:
return error on empty return code file (e.g. when application runs out of disk space)


emiloslavsky:
@jdidion this is what I figured out so far.  Please add the testcase in the ticket to dxCompiler test suite.

PR #162, opened 2021-08-06
APPS-751 Readme / script for extracting source code from metadata

r-i-v-a:


r-i-v-a:
Please review @jdidion

jdidion:
Is `base64` available on every platform, or do we have to list that as a prerequisite?

r-i-v-a:
> Is `base64` available on every platform, or do we have to list that as a prerequisite?

I think jq is also a prerequisite that needs to be installed; gunzip, sed, awk should be available I think? 

Do you have a specific place in mind for listing prerequisites, or do you just mean mention it in a comment in the script? @jdidion

r-i-v-a:
> I would mention it both in the script and in the `Internals` documentation.

Done, ready for review @jdidion

r-i-v-a:
Ready for review @jdidion

jdidion (approved):

PR #177, opened 2021-09-01
APPS-782: provide option to disable compile-time instance type evaluation

jdidion:
* add -instanceTypeSelection and -defaultInstanceType options
* do not evaluate instance types or store instance type db in applet metadata when -instanceTypeSelection is set to dynamic


jdidion:
This PR already has the change to use rank-based selection

RELEASE_NOTES.md
```diff
@@ -3,6 +3,8 @@
 ## in develop
 
 * Docker image dependencies that are DNAnexus platform files are included in applet's bundledDepends
+* Adds `-instanceTypeSelection` compiler option to allow disabling compile-type instance type selection
```
r-i-v-a, line 6: Ok, so this would still be optional and the alternative would be to cache the instance type DB based on what we got from the rank API during compile?

r-i-v-a (approved):

RELEASE_NOTES.md
jdidion: yes - the default is "static" which means to attempt compile-time determination of instance types and to store the instance type database in the applet metadata

PR #183, opened 2021-09-10
APPS-788 fix project-wide reuse

jdidion:
Also improves `CompilerTest` to clean use a separate folder and clean it up at the end of the suite.

### DIFF
--- a/compiler/src/main/scala/dx/compiler/DxExecutableDirectory.scala
+++ b/compiler/src/main/scala/dx/compiler/DxExecutableDirectory.scala
@@ -75,13 +75,16 @@ case class DxExecutableDirectory(bundle: Bundle,
     val desc: Option[DxObjectDescribe] = Some(dxDesc)
   }
 
-  private def findExecutables(folder: Option[String]): Vector[(DxDataObject, DxObjectDescribe)] = {
+  private def findExecutables(
+      folder: Option[String] = None,
+      recurse: Boolean = false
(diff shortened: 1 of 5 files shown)


r-i-v-a (changes requested):
-- Some tests are failing related to reorg
-- If this change is only in dxCompiler code, would like to include it in the release

jdidion:
tests are passing now

compiler/src/main/scala/dx/compiler/DxExecutableDirectory.scala
```diff
@@ -75,13 +75,16 @@ case class DxExecutableDirectory(bundle: Bundle,
...
+      folder: Option[String] = None,
+      recurse: Boolean = false
+  ): Vector[(DxDataObject, DxObjectDescribe)] = {
     Vector("applet", "workflow")
       .flatMap { dxClass =>
         val constraints = DxFindDataObjectsConstraints(
             project = Some(project),
             folder = folder,
-            recurse = false,
+            recurse = recurse,
```
emiloslavsky, line 87: does recursion happen within the folder or in the entire project (which is what should happen)?

jdidion: Entire project, unless a folder is specified (which it is not when `-projectWideReuse` is set)

r-i-v-a (approved):

PR #186, opened 2021-09-17
Release 2.5.0

r-i-v-a:
I have tried `git merge origin/main -X ours` on this branch, but there are conflicts that can't be automatically resolved. Also, this strategy for merging to main means that a lot of old commits will always show up in the branch diff, making it hard to see what is part of the new release.

Maybe we should just make release-X.Y.Z branches off of develop and keep them? It would also mean not having to move the git tag from the release branch onto main at the end of a release? @jdidion @commandlinegirl


jdidion:
Fixed. There was really only one conflict due to a file being moved in one branch and deleted in another. I find the value of a main branch questionable. The biggest argument for it that I see is that it is protected so it's hard to delete main where as it's easy to delete tags and release branches.

jdidion (dismissed):

r-i-v-a:
> Fixed. There was really only one conflict due to a file being moved in one branch and deleted in another. I find the value of a main branch questionable. The biggest argument for it that I see is that it is protected so it's hard to delete main where as it's easy to delete tags and release branches.

It's possible to add a protection rule for a name pattern as well as individual branches, so you could protect all `release-*` branches. https://docs.github.com/en/enterprise-server@3.0/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule

In case we deleted a tag by accident, it would generally be the last commit on the same-named release branch, so not hard to find where to put it back.

The process of merging into main just makes me a bit nervous, since so much history shows up in the diff (the commit ids once applied on main don't match the ones in develop / release branch's history), so it's hard to review, and I don't want to make a mistake while resolving conflicts that changes the record of what was released. @jdidion  @commandlinegirl

r-i-v-a:
We should still keep main around & protected, since it records some past releases. @jdidion  @commandlinegirl

PR #204, opened 2021-10-26
APPS-869 add cwl tests

YuxinShi0423:
https://jira.internal.dnanexus.com/browse/APPS-869
Added CWL tests from:
- standard cromwell cwl tests
- selected workflows

Results are summarized in: 
https://docs.google.com/spreadsheets/d/1lazAcM62S-Zvx02HtrNwB1so5-QSRxecw3-CYXb7VCA/edit#gid=0


r-i-v-a:
Add the label `integration` (on right sidebar) to this PR once you are ready to have the integration tests run. @YuxinShi0423

README.md
```diff
@@ -125,7 +125,7 @@ dxCompiler uses [wdlTools](https://github.com/dnanexus/wdlTools), a parser that
 - [Advanced options](doc/ExpertOptions.md) explains additional compiler options
 - [Internals](doc/Internals.md) describes current compiler structure (_work in progress_)
 - [Tips](doc/Tips.md) examples for how to write good WDL code
-- A high-level [list of changes](doc/WdlVersionChanges.md) between WDL draft-2 and version 1.0
+- A high-level [list of changes]([doc/WdlVersionChanges.md](https://github.com/openwdl/wdl/blob/main/versions/Differences.md#draft-2-to-10)) between WDL draft-2 and version 1.0
```
commandlinegirl: This link seems not to work (formatting is wrong).

scripts/run_tests.py
```diff
@@ -451,6 +451,43 @@
...
+    "cwl_cache_between_workflows",
+    "cwl_cache_within_workflow",
+    "cwl_docker_size",
+    "cwl_dynamic_initial_workdir",
+    "cwl_expressionLib",
+    "cwl_format",
+    "cwl_format_url",
+    "cwl_glob_sort",
+    "cwl_hello",
+    "cwl_http_inputs", # invalid input
```
commandlinegirl: What do the comments mean here? They say what the tests check?

test/cwl_cromwell/InitialWorkDirRequirement/file_array_listing_input.yaml
```diff
@@ -0,0 +1 @@
+files: ["/Users/chrisl/Downloads/cwl/allRequirements.txt"]  # array of type "File"
```
commandlinegirl: This path is not valid - it points to `/Users/chrisl/`.. Or perhaps for this test any path will work - if so, then please change it to something like /my/path/to/file, to show it's a dummy.

commandlinegirl (changes requested):
Mostly minor points and additionally:

- Please remove (and add to .gitignore) test/cwl_cromwell/.DS_Store 
- What does 'selected workflows' you mention in the description mean?

I requested access to the doc you mention in the description.

r-i-v-a:
In the release workflow, this runs the large tests
https://github.com/dnanexus/dxCompiler/blob/develop/.github/workflows/release.yml#L114

You can temporarily edit integration.yml in your branch to run them here, too.
Revert eventually before merging.
https://github.com/dnanexus/dxCompiler/blob/develop/.github/workflows/integration.yml#L61

And also apply the `integration` label on this PR. @YuxinShi0423

r-i-v-a (changes requested):
Temporarily edit integration.yml so the most recent test results are visible here.

Schedule meeting next week if needed to fill in context, especially re: what tests had to be changed and what is the process to deal with tests that still fail. @YuxinShi0423

scripts/run_tests.py
r-i-v-a: I think they are annotations on currently failing tests (reason why they are failing). Should clarify how this will be handled before merging -- I see on other tests that @jdidion 's approach is to comment out the whole line, and leave a comment on the right about what fix is pending to make the test work.

.github/workflows/integration.yml
```diff
@@ -58,4 +58,10 @@ jobs:
           cd ${GITHUB_WORKSPACE}
           FOLDER=/builds/test_gha_dxcompiler_$(date +%s)_$RANDOM
           dx mkdir $FOLDER
-          python3 scripts/run_tests.py --test M --folder $FOLDER --project $PROJECT
+          TEST=cwl_cromwell
+          python3 scripts/run_tests.py --test $TEST --folder $FOLDER --project $PROJECT --delay-compile-errors --delay-run-errors
+      - name: Upload artifact
```
r-i-v-a: This "Upload artifact" step should be removed when you update the branch with develop. It has been removed in develop.

scripts/run_tests.py
r-i-v-a: To do: plan for fixing these on a case-by-case basis? They look like they are failing for a variety of reasons. E.g. "Could not resolve host: metadata.google.internal" is probably something we don't have access to.

scripts/run_tests.py
```diff
@@ -1325,7 +1368,8 @@ def run_test_subset(
 
     if errors:
         write_failed(errors)
-        raise RuntimeError(f"failed to run one or more tests {','.join(errors)}")
```
r-i-v-a: Shouldn't this be reverted?

scripts/run_tests.py
```diff
@@ -1683,11 +1727,15 @@ def compile_tests_to_project(
...
                     raise
-        runnable[tname] = oid
-        print("runnable({}) = {}".format(tname, oid))
+        if oid:
+            runnable[tname] = oid
+            print("runnable({}) = {}".format(tname, oid))
     if errors:
         write_failed(errors)
-        raise RuntimeError(f"failed to compile one or more tests: {','.join(errors)}")
+        cprint(f"failed to compile one or more tests: {','.join(errors)}", "red")
```
r-i-v-a: Shouldn't this be reverted?

r-i-v-a:
> * Please remove (and add to .gitignore) test/cwl_cromwell/.DS_Store

After removing those files, you can add this to .gitignore

```
# macOS
**/.DS_Store
```

That will ignore it in any folder.

test/cwl_cromwell/InitialWorkDirRequirement/file_array_listing_input.yaml
r-i-v-a: I guess this depends on our approach to using the Cromwell test suite. Do we want to leave the tests "as-is" as much as possible (assuming this passes as-is) or make our own changes?

Related -- when a Cromwell test fails due to what we think is a problem in the test suite (not a problem in our implementation), do we want to send a fix PR back to them? @commandlinegirl  @jdidion

r-i-v-a:
-- What does y in column C mean? That the test compiled _and_ ran successfully if cwl-upgrader was applied? Does it mean that cwl-upgrader had to be applied, otherwise it did not compile / run at all?

-- What does y in column E mean? Does it mean that validation passed? On the original test CWL or after applying upgrade? Why do many tests fail in column E but pass in column C?

r-i-v-a:
-- Or columns C through G indicate whether it was possible to apply the tool to the test CWL, column I is whether dxCompiler was able to compile the workflow, and J is whether the workflow also ran successfully?

r-i-v-a (changes requested):
To discuss at meeting.

.github/workflows/integration.yml
YuxinShi0423: Fixed.

scripts/run_tests.py
YuxinShi0423: Reverted.

scripts/run_tests.py
YuxinShi0423: Reverted. Thanks :D

scripts/run_tests.py
```diff
@@ -452,6 +452,45 @@
...
 
+cwl_cromwell_tests_list = [
+    "cwl_ad_hoc_file_test",
+    "cwl_cache_between_workflows",
+    "cwl_cache_within_workflow",
+    "cwl_docker_size",
+    "cwl_dynamic_initial_workdir",
+    "cwl_expressionLib",
+    "cwl_format",
+    # "cwl_format_url", # Could not load extension schema https
```
r-i-v-a: For ignored tests, please reference the Jira id where the issue will be resolved, so we know when to un-ignore them. @YuxinShi0423 @jdidion

r-i-v-a:
I have just added the `integration` label since it should be run on all non-trivial PRs, but since the `large` suite isn't included in that, please just confirm via comment that all currently non-ignored tests are passing (e.g. when you tested it via CLI). @YuxinShi0423

r-i-v-a (changes requested):
Update branch with `develop` & see comments.

YuxinShi0423:
https://github.com/dnanexus/dxCompiler/pull/267
Due to compilation failure using sbt 1.5.7 the integration test cannot be run. @commandlinegirl @r-i-v-a I will work on this PR as soon as we get it fixed. Thanks!

r-i-v-a (dismissed):

YuxinShi0423:
> Mostly minor points and additionally:
> 
> * Please remove (and add to .gitignore) test/cwl_cromwell/.DS_Store
> * What does 'selected workflows' you mention in the description mean?
> 
> I requested access to the doc you mention in the description.

@commandlinegirl 
.DS_Store has been removed. Would you mind reviewing this PR again?

commandlinegirl (approved):
LGTM

PR #214, opened 2021-11-08
put reserved parameters in a group

jdidion:


RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@
 ## in develop
 
 * Handles empty scatters in WDL workflows
+* Reserved parameters are now placed in the "Reserved" parameter group (only affects the display of the app/workflow in the UI)
```
commandlinegirl: Should we name the group something that is very clear to the user that it's something system-like, set by `dxCompiler`? E.g. `Reserved for dxCompiler (do not change)` or just `Reserved for dxCompiler`.

compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
```diff
@@ -23,33 +23,46 @@ import dx.translator.CallableAttributes.{
...
 }
+import dx.translator.ParameterAttributes.GroupAttribute
 import dx.translator.{Extras, ParameterAttributes}
 import dx.util.FileSourceResolver
 import spray.json._
 
 object ExecutableCompiler {
-  // these parameters are used for applets and workflows that are generated
-  // with useManifests=true.
+  // these parameters are used for applets and workflows that are generated with useManifests=true.
```
commandlinegirl, line 32: Would it be possible to add a short comment on why the overrides are needed?

r-i-v-a (changes requested):
What Ola said about adding comment.

r-i-v-a (dismissed):

PR #215, opened 2021-11-08
APPS-907 add expression values to output manifests

jdidion:
If a call or block is preceded by private variables, the evaluation of those variables is bundled together with the call/block in the call graph (using a helper applet). When using manifests, the helper applet just returns a link to the output manifest of the call/block. However, downstream expressions may reference the expressions evaluated by the helper app, so we need the evaluated expressions to go into the call/block's output manifest. This PR adds another input parameter to generated applets (`extra_outputs___`) that takes a hash and simply appends it to the outputs that get serialized into the output manifest.

This accidentally got branched off of APPS-899, so wait to merge this until that one is merged.


compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
```diff
@@ -44,6 +44,8 @@ object ExecutableCompiler {
     Parameter(Constants.WorkflowInputLinks, Type.TOptional(Type.THash))
   val OutputIdParameter: Parameter =
     Parameter(Constants.OutputId, Type.TString)
+  val ExtraOutputsParameter: Parameter =
```
r-i-v-a, line 47: Shouldn't this be combined with the changes in https://github.com/dnanexus/dxCompiler/pull/214/files -- reserved parameter group? @jdidion

r-i-v-a (changes requested):
Merge reserved parameter group changes first, then apply that as needed here.

compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
jdidion: They are separate changes. I'll take care of it when I merge the other PR.

r-i-v-a (approved):

PR #217, opened 2021-11-09
APPS-910 Pre-emptively install compatible version of pyparsing before cwltools.

r-i-v-a:


r-i-v-a:
Even if it works, try re-running the test jobs a couple times to check consistency. @jdidion

jdidion (approved):

PR #218, opened 2021-11-10
APPS-747 update ExpertOptions doc

YuxinShi0423:


doc/ExpertOptions.md
```diff
@@ -43,33 +43,45 @@ Make sure you've installed the dx-toolkit CLI, and initialized it with `dx login
...
+| execTree \[json,pretty\] | Print a JSON representation of the workflow. |
+| extras \<string\> | JSON file with extra options (see documentation). |
+| inputs \<string\> | JSON file with standard-formatted input values. May be specified multiple times. A DNAnexus JSON input file is generated for each standard input file. |
+| instanceTypeSelection \[static,dynamic\] |  Whether to attempt to select instance types at compile time or tasks with runtime requirements that can all be statically valuated (static, the default), or to defer all instance type election to runtime (dynamic). Using static instance type election can save time and cost, but it requires the user(s) unning the applet/workflow to have access to the same instance ypes as the user who compiled it.
+| imports \<string\> | Directory to search for imported WDL files. May be specified multiple times. |
+| locked   | Create a locked workflow. When running a locked workflow, input values may only be specified for the top-level workflow. |
+| leaveWorkflowsOpen | Leave created workflows open (otherwise they are closed). |
 | noListings | CWL-specific option to prevent full folder listings in generated input files |
```
YuxinShi0423: the `noListings` option doesn't seem to be implemented. However it is described in Line 250. Is there anything I had missed? @jdidion @r-i-v-a

r-i-v-a: I would just copy here what is currently in compiler/src/main/scala/dxCompiler/Main.scala usageMessage on `develop` branch, which does not currently include noListings.

But I see it is also mentioned here
"CWL does provide a mechanism for ensuring reproducibility of jobs that take directory inputs, via the `listing` field. We strongly recommend that CWL users specify the folder listing for each directory input. A job will only be reused if both the folder and the listing are identical. The ordering of the listing is taken into consideration when making the comparison, so the listing must be generated deterministically. The default behavior of dxCompiler when using the `-input` option is to generate input files with full listings for all directories, unless the `-noListings` option is specified. An example of a folder with a listing is:"

I think it should not be mentioned in the documentation if not implemented yet; once it is implemented, we can update both the usageMessage and the docs. @jdidion

r-i-v-a (changes requested):
Question for @jdidion  about status of `noListings`. 
For now, I think the docs should just be copied from current `usageMessage` in Main.scala. @YuxinShi0423

jdidion:
@r-i-v-a You're right, I had planned to implement that but didn't get to it. We can remove it from the docs for now and add a Jira issue to implement it.

r-i-v-a:
> @r-i-v-a You're right, I had planned to implement that but didn't get to it. We can remove it from the docs for now and add a Jira issue to implement it.

https://jira.internal.dnanexus.com/browse/APPS-914 @jdidion @YuxinShi0423

doc/ExpertOptions.md
```diff
@@ -623,8 +635,10 @@ The WDL definition file will be:
 
 ```wdl
 task concat {
-  String a
-  String b
+  input {
```
r-i-v-a: this looks fine @YuxinShi0423

r-i-v-a (approved):

PR #220, opened 2021-11-10
Put `extra_outputs__` and `output_id___` in reserved group

jdidion:


compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
```diff
@@ -29,43 +29,90 @@ import dx.util.FileSourceResolver
 import spray.json._
 
 object ExecutableCompiler {
-  // these parameters are used for applets and workflows that are generated with useManifests=true.
-  val ReservedParameterAttributes = Vector(GroupAttribute("Reserved for dxCompiler"))
+  // Parameters used for applets and workflows that are generated with useManifests=true.
```
commandlinegirl, line 32: Has this file got committed by mistake?

jdidion: No

commandlinegirl (approved):

compiler/src/main/scala/dx/compiler/ExecutableCompiler.scala
commandlinegirl: I thought this is only about GHA when I got here from gchat. Please add a note that thus commit includes the minor label in the commit message

PR #223, opened 2021-11-11
APPS-916 Allow native app(let) to be specified in runtime section using either id or name.

jdidion:
Fixes a bug in current implementation and adds support for the `dx_app` key in the `runtime` section (pre-WDL 2.0). This is needed by SciProd for workflow development.


commandlinegirl:
@jdidion Could you add a JIRA ticket for this if there isn't one?

r-i-v-a:
> @jdidion Could you add a JIRA ticket for this if there isn't one?

Created https://jira.internal.dnanexus.com/browse/APPS-916

r-i-v-a (approved):

r-i-v-a:
I think @jkotrs  also wanted the ability to specify a single app to generate the WDL wrapper for when calling `dxni`. This would save time on generating wrappers for all visible apps when only 1 is needed, and help when `dxni` would fail for a different app before reaching the app that was wanted. 

Do we have a Jira for that -- @jkotrs might have created one already? @jdidion

PR #225, opened 2021-11-11
APPS-804 Make runtime instance type selection the default. [WIP]

r-i-v-a:


executorCommon/src/main/scala/dx/executor/JobMeta.scala
```diff
@@ -757,14 +757,7 @@ abstract class JobMeta(val workerPaths: DxWorkerPaths,
       InstanceType.createDb(dxApi = dxApi)
     } catch {
       case _: Throwable =>
-        // fall back to the cached database if available
```
r-i-v-a, line 760: If we no longer cache the instanceTypeDb, we rely on fetching it for the project at runtime (above) succeeding.

With global WDL workflows / clonable workflows, we can't make assumptions about the future projects a workflow can be run in at compile time, so I think not caching instanceTypeDb is correct.

I also think a user who has the ability to run workflows in a project (CONTRIBUTE access) would also have the permissions for the above call, or are there cases where we could reach this error? Instance type selection should now be based on the pricing rank, so should also be accessible to any user with CONTRIBUTE access, correct? @jdidion

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -166,7 +167,14 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
         instanceTypeDb.apply(static.toInstanceTypeRequest).name
       case DefaultInstanceType | DynamicInstanceType =>
         // TODO: should we use the project default here rather than
```
r-i-v-a, line 168: What is meant by this -- I can resolve this todo in the PR if you want? @jdidion

r-i-v-a:
Closing due to change of spec.

PR #230, opened 2021-11-15
Native stub generator produces runtime -> dx_app object 

jkotrs:
with name, version and id


jdidion (dismissed):

jdidion (approved):

emiloslavsky:
@jdidion  - why is this change introduced?

PR #233, opened 2021-11-16
APPS-930 fix instance type selection logic for calling native apps

jdidion:
Previously, the code generator was generating the wrong meta section for native apps (`"type": true`). Prior to 2.6, the code in the workflow fragment executor that determines whether the app it is executing is native accounted for that, but in 2.7 the code was updated to look for the expected value (`"type": "native"`), which led to a regression. This PR fixes the code generator to produce the correct meta section.

I also updated the release workflow to run the unit tests and cleaned up some formatting.


.github/workflows/release.yml
```diff
@@ -69,7 +69,31 @@ jobs:
...
+          pip3 install termcolor dxpy $cwltool
+          export PATH="$PATH:$HOME/.local/bin"
+
+          # install dxda
+          sudo wget -O/usr/local/bin/dx-download-agent https://github.com/dnanexus/dxda/releases/download/v0.5.7/dx-download-agent-linux
+          sudo chmod +x /usr/local/bin/dx-download-agent
+
+          # set up DNAnexus staging environment
+          dx login --noprojects --staging --token $AUTH_TOKEN
+          dx select dxCompiler_playground
```
commandlinegirl: How about including an `env` section to the workflow with environment variable set to dxCompiler_playground (since it's used in a step below as well - line 106)

https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables

test-dxcompiler-in-project/replace_dxcompiler_jar.sh
```diff
@@ -0,0 +1,11 @@
+#/bin/bash
```
commandlinegirl: This script was removed in https://github.com/dnanexus/dxCompiler/pull/228/files. Why reintroduce it?

commandlinegirl (changes requested):

test-dxcompiler-in-project/replace_dxcompiler_jar.sh
jdidion: I guess it was a merge error

.github/workflows/release.yml
jdidion: done

commandlinegirl (approved):
LGTM, thanks!

PR #237, opened 2021-11-18
Use manifest files for large subjob inputs

jdidion:
Previously, when using manifests, we passed the manifest to a subjob as a hash. In some cases, the subjob input can still be very large. This PR checks the size of the input JSON document, and if it is larger than a set limit, a manifest file is used instead.

### DIFF
--- a/executorCommon/src/main/scala/dx/executor/JobMeta.scala
+++ b/executorCommon/src/main/scala/dx/executor/JobMeta.scala
@@ -544,8 +544,8 @@ abstract class JobMeta(val workerPaths: DxWorkerPaths,
         // files, so put them in a subfolder.
         val filename =
           s"${jobId}_subjob${nameDetail.map(d => s"/${d}").getOrElse("")}.manifest.json"
-        val destination = s"${manifestProjectAndFolder}/${filename}"
-        val manifestDxFile = dxApi.uploadString(manifestJsStr, destination)
+        val manifestDxFile =
+          dxApi.uploadString(manifestJsStr, s"${manifestProjectAndFolder}/${filename}")
         Map(Constants.InputManifestFiles -> JsArray(manifestDxFile.asJson))
       }
       val commonInputsJs = Vector(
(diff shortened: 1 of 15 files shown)


executorCommon/src/main/scala/dx/executor/WorkflowExecutor.scala
```diff
@@ -181,9 +181,11 @@ abstract class WorkflowExecutor[B <: Block[B]](jobMeta: JobMeta, separateOutputs
 
     val prefix = if (prefixOutputs) Some(name) else None
     val callInputsJs = JsObject(
-        jobMeta.prepareSubjobInputs(inputs, extraManifestOutputs, executableLink, prefix).map {
-          case (dxName, jsv) => dxName.encoded -> jsv
-        }
+        jobMeta
+          .prepareSubjobInputs(inputs, extraManifestOutputs, executableLink, prefix, nameDeail)
```
commandlinegirl:
```suggestion
          .prepareSubjobInputs(inputs, extraManifestOutputs, executableLink, prefix, nameDetail)
```

executorCommon/src/main/scala/dx/executor/JobMeta.scala
```diff
@@ -55,6 +56,7 @@ object JobMeta {
   // this file has all the information about the executable
   val ExecutableInfoFile = "dnanexus-executable.json"
   val MaxConcurrentUploads = 8
+  val MaxManifestJsLength = 100_000
```
commandlinegirl, line 60: Could you add a comment on what this length is (number of chars)?

commandlinegirl (changes requested):

executorCommon/src/main/scala/dx/executor/JobMeta.scala
jdidion: done

commandlinegirl (approved):

PR #239, opened 2021-11-19
APPS-937 Use dxda to download multiple manifest files

jdidion:
When there are multiple manifest files to download, uses dxda to avoid API throttling.


compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -135,7 +135,8 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
             ApplicationCompiler.WorkflowFragmentTemplate,
             templateAttrs ++ Map(
                 "separateOutputs" -> separateOutputs,
-                "waitOnUpload" -> waitOnUpload
+                "waitOnUpload" -> waitOnUpload,
+                "dxPathConfig" -> runtimePathConfig
```
commandlinegirl: Why not, for consistency and to meet the expectation of the developer reading the code, use the dame name for the value, i.e. `dxPathConfig` (or use `runtimePathConfig` in both key and value)?

compiler/src/main/resources/templates/workflow_script.ssp
```diff
@@ -0,0 +1,65 @@
...
+              echo "The download log is:"
+              cat "${dxPathConfig.getDxdaWorkflowManifestDownloadManifestFile().toString}.bz2.download.log"
+           fi
+           exit ${bashDollar}dxda_error_code
+       fi
+
+       # The download was ok, check file integrity on disk
+       dx-download-agent inspect "${dxPathConfig.getDxdaWorkflowManifestDownloadManifestFile().toString}.bz2"
+
+       # go back to root directory
```
commandlinegirl: Don't we need to go back to the dir we left in line 45?

executorCommon/src/main/scala/dx/executor/JobMeta.scala
```diff
@@ -58,6 +60,9 @@ object JobMeta {
   val MaxConcurrentUploads = 8
   // max subjob input manifest document size for which a hash is used
   val MaxManifestJsLength = 100_000
+  // functions used to download manifest files
+  val ManifestDownloadDxda = "manifest_download_dxda"
```
commandlinegirl, line 64: What is the difference between manifest and workflow manifest?

commandlinegirl (changes requested):
- Could you make sure the JIRA ticket ID is set on the PR and the commit message (JIRA IDs in commit messages makes it easier to check if a particular commit is in production (release branch) or not.
- Could you add a test for this?

executorCommon/src/main/scala/dx/executor/JobMeta.scala
jdidion: Two different sets of files downloading to two different directories

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
jdidion: done

compiler/src/main/resources/templates/workflow_script.ssp
jdidion: the cd is unnecessary, I removed it

jdidion:
There are already tests

commandlinegirl (approved):

PR #240, opened 2021-11-19
APPS-993 capture all stage prefixes when parsing WdlDxName

jdidion:


r-i-v-a (approved):

core/src/main/scala/dx/core/ir/DxName.scala
```diff
@@ -288,7 +288,7 @@ trait DxNameFactory {
 object DxNameFactory {
   // standard suffixes to parse
   val suffixes = Set(Constants.ComplexValueKey, Constants.FlatFilesSuffix)
-  private val stageRegex = "^(?:(stage-[^.]+)\\.)?(.+)$".r
+  private val stageRegex = "^(?:((?:stage-[^.]\\.)*stage-[^.]+)\\.)?(.+)$".r
```
emiloslavsky, line 291: please add a testcase

PR #241, opened 2021-11-19
APPS-936 File to string

jdidion:
In the case where an input has type string and a user passes a dx uri string, the uri is automatically converted to a file (`VFile`). There is then an error when trying to coerce the file to string. This PR allows the file-to-string coercion.


commandlinegirl (changes requested):
Thank you for the fix! Could you add a unit test for this?

jdidion:
done

commandlinegirl (dismissed):
Thank you!

PR #243, opened 2021-11-22
APPS-943 add task to run customer workflows

jdidion:
This is the proposed changes to the release GHA to (optionally) run customer workflows. This will require:

* GH secrets with the customer project-scoped token, project ID, and workflow names.
* One app for each workflow that takes a workflow ID, and runs the workflow. We do this rather than launch the workflow directly from the GHA so we can use `dx watch` to wait on the run completion and then fail if the workflow failed.


.github/workflows/release.yml
```diff
@@ -130,6 +135,45 @@ jobs:
...
+          export PATH="$PATH:$HOME/.local/bin"
+
+          # set up DNAnexus production environment for customer 1
+          dx login --noprojects --token CUSTOMER1_AUTH_TOKEN
+          dx select $CUSTOMER1_PROJECT
+
+          # download, compile, and run workflows
+          dx download -r /workflow_1
+          cd workflow_1
+          $workflow1=$(java -jar ./dxCompiler-${{ github.event.inputs.release-version }}.jar compile \
```
r-i-v-a: How does the dxCompiler .jar exist -- from being compiled in an earlier GHA step?

.github/workflows/release.yml
```diff
@@ -130,6 +135,45 @@ jobs:
               --docker-password $DOCKERHUB_TOKEN \
               --force \
               --branch $TARGET_BRANCH
+      - name: Run customer 1 workflows
+        if: ${{ github.event.inputs.should-run-customer-tests == 'true' }}
+        env:
+          CUSTOMER1_AUTH_TOKEN: ${{ secrets.CUSTOMER1_TOKEN }}
```
r-i-v-a, line 141: Did we check with the relevant people to make sure it's ok to have this in GHA?

How do we plan to generate this token -- a limited-access token from one of our accounts that is in the org that has access to the projects?

Where will we track (internally) which customer / project is "customer1" "customer2" etc. and for which projects the customer token needs CONTRIBUTE, VIEW, etc. for when we have to re-generate the tokens in the future? @jdidion  @commandlinegirl

test-dxcompiler-in-project/Makefile
```diff
@@ -1,6 +1,6 @@
 # TODO: find a way to download from draft release URLs; the below works for final releases.
```
r-i-v-a, line 1: With this setup, we could remove the test-dxcompiler-in-project applet, since it wouldn't be used.

.github/workflows/release.yml
r-i-v-a: Why not put compilation outputs and run outputs into a folder with the timestamp, the way the current script does? That makes it easier to debug a specific run, or clean up the output from older runs.

jdidion: 1) yes
2) done, thanks

.github/workflows/release.yml
jdidion: All good questions. I'm open to suggestions. My proposal is to manage it in a Google sheet. For tokens, either the customer can add us to their project with proper permissions and we can generate the scoped token, or they can generate the token and share it with us securely. The former is preferable so we don't have to bug the customer every time the token needs to be refreshed.

test-dxcompiler-in-project/Makefile
jdidion: correct

PR #245, opened 2021-11-24
APPS-944 Remove unused upload artifact step from GHA, try shorter release WF name.

r-i-v-a:


.github/workflows/integration.yml
```diff
@@ -59,8 +59,3 @@ jobs:
           FOLDER=/builds/test_gha_dxcompiler_$(date +%s)_$RANDOM
           dx mkdir $FOLDER
           python3 scripts/run_tests.py --test M --folder $FOLDER --project $PROJECT
-      - name: Upload artifact
```
r-i-v-a, line 62: @commandlinegirl  previously used this for running some tests manually, but it isn't needed anymore. @jdidion

.github/workflows/release.yml
```diff
@@ -1,4 +1,4 @@
-name: dxCompiler Release (Staging and Prod)
+name: Release dxCompiler
```
r-i-v-a, line 1: I think the long name is why it isn't displaying correctly in the left panel here? 
https://github.com/dnanexus/dxCompiler/actions

commandlinegirl (approved):
LGTM

PR #250, opened 2021-11-29
APPS-956 Localize private variables.

jdidion:
@r-i-v-a this should fix the issue with private variables not being localized. Needs a test though.


r-i-v-a:
Added test case, running integration tests now.

r-i-v-a (approved):
I added a test and it passes.

PR #253, opened 2021-12-03
allow runtime of native stub to override default instance type of native app(let)

jdidion:
If the WDL task stub for a native app(let) specifies runtime information, that information should be used to determine the instance type. Note that this only handles the cases where the native app(let) is called from a fragment or where the instance type is static.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 
 * Excludes apps from `bundledDepends`
 * Localizes files declared in WDL task private variables
+* Respects runtime definition in native task stub
 
 ## 2.8.0 2021-11-29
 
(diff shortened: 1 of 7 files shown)


r-i-v-a (approved):

emiloslavsky:
Added https://jira.internal.dnanexus.com/browse/APPS-1002 to update dxCompiler documentation and tests to reflect native stub to override default instance type of native app(let).
Not clear if the description in this pull request `If the WDL task stub for a native app(let) specifies runtime information, that information should be used to determine the instance type. Note that this only handles the cases where the native app(let) is called from a fragment or where the instance type is static.` or the 
https://jira.internal.dnanexus.com/browse/PMUX-904?focusedCommentId=176418&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-176418 applies

emiloslavsky:
@commandlinegirl @r-i-v-a @jdidion FYI

PR #254, opened 2021-12-07
APPS-627 Add e2e test for global workflow from WDL.

r-i-v-a:


r-i-v-a:
This is ready for an initial review -- what is still to be done is revert to installing dxpy normally after the fixed version is released. Right now I install dxpy from a git ref in order to test against Yuxin's branch. @YuxinShi0423 @jdidion @commandlinegirl

r-i-v-a:
Ran my e2e test against Yuxin's branch

https://github.com/dnanexus/dx-toolkit/pull/753
https://github.com/dnanexus/dxCompiler/runs/4870350980?check_suite_focus=true

Confirmed fixes for APPS-975 and APPS-1028. @YuxinShi0423

r-i-v-a:
Integration tests pass using the recent dxpy release. @emiloslavsky @YuxinShi0423

.github/workflows/integration.yml
```diff
@@ -42,21 +42,28 @@ jobs:
...
-          python3 scripts/run_tests.py --test M --folder $FOLDER --project $PROJECT
+
+          # Run tests
+          python3 scripts/run_tests.py --folder $FOLDER --project $PROJECT \
+          --test M
+          python3 scripts/run_multiuser_tests.py --folder $FOLDER --project $PROJECT \
+          --alice-token $ROBOT_TOKEN \
+          --bob-token $TEST_ROBOT_TOKEN
+
+          # TODO Also ensure folder is removed when tests fail
```
YuxinShi0423: Any reason why it is still a TODO?

scripts/util.py
```diff
@@ -459,6 +459,7 @@ def build_executable(source_file, project, folder, top_dir, version_id, compiler
 # oid                           Id of executable to run
```
YuxinShi0423: Minor: id or name

scripts/run_multiuser_tests.py
```diff
@@ -0,0 +1,303 @@
...
+    register_project(project)
+
+    # Do folder setup as Alice
+    login_alice()
+
+    if args.folder is None:
+        base_folder = util.create_build_dirs(project, version_id)
+    else:
+        # Use existing prebuilt base folder
+        base_folder = args.folder
```
YuxinShi0423, line 280: trivia: it allows us to use assets of a different version inside the args.folder instead the ones of `version_id`.

scripts/run_multiuser_tests.py
```diff
@@ -0,0 +1,303 @@
...
+        ))
+        raise
+
+    # As Bob, run global workflow
+    login_bob()
+
+    exec_objs = util.run_executable(
+        oid=global_workflow_name,
+        project=PROJECT,
+        test_folder=specific_applet_folder(tname),
```
YuxinShi0423: trivia: dest folder might be specific_test_folder

.github/workflows/release.yml
```diff
@@ -100,24 +100,31 @@ jobs:
...
-          #TODO: also remove the folder when the tests fail
+
+          # Run tests
+          python3 scripts/run_tests.py --folder $FOLDER --project $PROJECT \
+          --test L
+          python3 scripts/run_multiuser_tests.py --folder $FOLDER --project $PROJECT \
+          --alice-token $ROBOT_TOKEN \
+          --bob-token $TEST_ROBOT_TOKEN
+
+          # TODO Also ensure folder is removed when tests fail
```
YuxinShi0423: same here: TODO?

YuxinShi0423 (dismissed):
LGTM. Thank you!

.github/workflows/integration.yml
r-i-v-a: It was a TODO (at least in release.yml) before this PR, and I don't make any change to that implementation here. I marked as TODO in both workflows because I don't see a difference in implementation.

.github/workflows/release.yml
r-i-v-a: This was originally a TODO before this PR, and I didn't change the implementation.

scripts/util.py
r-i-v-a: Edited

scripts/run_multiuser_tests.py
r-i-v-a: I edited the documentation for the folder arg and the comment, in both test scripts, to "Use an existing folder with dxCompiler assets, instead of building dxCompiler."

r-i-v-a: I don't think this can be used with another version, since version_id is used in the method to compile workflows and that would probably not work when it doesn't match. It's probably just to avoid compiling the same code again (but for GHA, we always have to do a fresh compilation on the worker).

scripts/run_multiuser_tests.py
r-i-v-a: Good point, now edited to use specific_test_folder.

.github/workflows/integration.yml
r-i-v-a: Changed to running each script with `--clean`, which should remove the folder in the finally block even if some tests failed.

.github/workflows/release.yml
```diff
@@ -100,24 +100,34 @@ jobs:
...
+          ROBOT_TOKEN: ${{ secrets.DX_STAGING_ROBOT_TOKEN }}
+          TEST_ROBOT_TOKEN: ${{ secrets.DX_STAGING_TEST_ROBOT_TOKEN }}
         run: |
           export PATH="$PATH:$HOME/.local/bin"
 
-          # set up DNAnexus staging environment
-          dx login --noprojects --staging --token $AUTH_TOKEN
+          # Set up DNAnexus staging environment
+          dx login --noprojects --staging --token $ROBOT_TOKEN
+          PROJECT=dxCompiler_playground
```
commandlinegirl: PROJECT env variable doesn't have to be specified here since [it set for the GHA workflow already ](https://github.com/dnanexus/dxCompiler/blob/e041c572057a4ecad38da7b11e903059ddccb778/.github/workflows/release.yml#L20). For consistency, if you could set such variable in `integration.yml` as well.

commandlinegirl (changes requested):
Only a minor thing found.

.github/workflows/release.yml
r-i-v-a: Done

commandlinegirl (approved):
LGTM, thanks!

PR #255, opened 2021-12-07
Fix executable archiving

jdidion:
Currently, when the compiler is run with the `-a` flag, any existing executables with name collisions are archived by renaming them. This breaks the archived workflow because in other parts of the code we rely on the executable name being the same as the task name. This PR fixes the archival process to instead move the executable to a `.archive/<name> <date>` folder and adds the "archived" tag. We then exclude executables with the "archived" tag when searching.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 
 * Excludes apps from `bundledDepends`
 * Localizes files declared in WDL task private variables
+* Fixes archiving of executables - tags them as "archived" rather than renaming them
 
 ## 2.8.0 2021-11-29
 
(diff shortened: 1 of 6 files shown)


emiloslavsky:
Can we use `dxCompilerArchived` tag instead of `archived` for disambiguation?

RELEASE_NOTES.md
```diff
@@ -5,6 +5,7 @@
 * Excludes apps from `bundledDepends`
 * Localizes files declared in WDL task private variables
 * Respects runtime definition in native task stub
+* Fixes archiving of executables - tags them as "archived" rather than renaming them
```
r-i-v-a: Edit release notes to actual tag name
```
val ArchivedTag = "dxCompilerArchived"
```

r-i-v-a (dismissed):
Edit release notes entry, otherwise ok.

PR #260, opened 2021-12-09
fix regex for parsing stage IDs

jdidion:
Fix for regex that misses stage IDs >= 10


commandlinegirl (changes requested):
Would you add unit tests for checking the ID?

r-i-v-a (approved):

PR #265, opened 2021-12-15
APPS-1004 Update Scala projects to sbt v1.5.7.

r-i-v-a:

PR #267, opened 2021-12-17
APPS-1004 Branch for debugging / testing sbt upgrade.

r-i-v-a:
Branch for debugging / testing sbt upgrade.

Revert "APPS-1004 Update Scala projects to sbt v1.5.7. (#265)"
This reverts commit d90a8f8cfba66d282459a757441b725f4094ea8d. @jdidion  @commandlinegirl

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,5 @@
 import Merging.customMergeStrategy
 import sbt.Keys._
-import sbtassembly.AssemblyPlugin.autoImport.{assemblyMergeStrategy, _}
 import com.typesafe.config._
 
 name := "dxc"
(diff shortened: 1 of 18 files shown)


build.sbt
```diff
@@ -33,7 +33,7 @@ lazy val global = root
       settings,
       publish / skip := true
   )
-  .disablePlugins(AssemblyPlugin)
+//  .disablePlugins(AssemblyPlugin)
```
r-i-v-a: This line wasn't changed recently, but why was it set this way? (Getting rid of it doesn't seem to help with the current issue.) @jdidion

project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.1.0")
```
r-i-v-a, line 1: From https://github.com/sbt/sbt-assembly/releases

build.sbt
jdidion: You need to skip assembly on the global project or you'll get a basically empty JAR created for the root project.

r-i-v-a: Aha, thanks @jdidion

jdidion (dismissed):

jdidion (approved):

PR #278, opened 2022-01-08
Apps-897 switch from cwltool to cwlpack

jdidion:


core/src/main/scala/dx/core/languages/cwl/CwlBlock.scala
```diff
@@ -349,46 +354,47 @@ object CwlBlock {
...
       if (satisfied.isEmpty) {
         throw new Exception("could not satisfy any dependencies")
       }
-      val newDeps = deps ++ satisfied.map(step => step.name -> step).to(SeqMap)
+      val newDeps = deps ++ satisfied.map(step => step.frag -> step).to(SeqMap)
       if (unsatisfied.isEmpty) {
         newDeps
       } else {
-        newDeps ++ orderSteps(unsatisfied, newDeps)
+        newDeps ++ orderSteps(wfId, unsatisfied, newDeps)
```
YuxinShi0423, line 382: @jdidion when unsatisfied is not empty, we call the orderSteps again, but those steps will still be recognized as unsatisfied this time, right?

r-i-v-a: `deps = newDeps` will have added contents since the last call, so there is the chance that in each successive call, some more items will get partitioned into `satisfied` via 

```
case id if id.parent.exists(deps.contains) => true
```

YuxinShi0423: Oh just realized it is not looping within one step. Thanks Riva!

jdidion: no, because the dependencies have been updated

r-i-v-a (approved):

r-i-v-a (changes requested):
Please add to release notes and fix what looks like repeated section at the top.

r-i-v-a (dismissed):
Ignore earlier comment, I think there was a glitch in the GitHub review UI -- it is having problems handling all these files.

r-i-v-a (dismissed):

r-i-v-a (approved):

r-i-v-a:
I experienced some GitHub glitch while trying to merge this -- after clicking "Squash and merge," it stalled and didn't close the PR.

However, it looks like it's now successfully applied on develop -- there are 2 commits with this name, the first contains the actual contents from this PR and the second is empty.
-- https://github.com/dnanexus/dxCompiler/commit/bae9172ee010098c3f997ba476326938b1b59722
-- https://github.com/dnanexus/dxCompiler/commit/5915f947be1c5cce9c8105208b2657ecaf366ad9

PR #284, opened 2022-01-13
APPS-1022 Fix for using step default in workflow cwl

YuxinShi0423:

PR #285, opened 2022-01-13
APPS-1022 get defaults from step inputs

jdidion:


r-i-v-a:
Being worked on in https://github.com/dnanexus/dxCompiler/pull/284

PR #286, opened 2022-01-14
APPS-1024 update scalafmt and sbtscalafmt

YuxinShi0423:


RELEASE_NOTES.md
```diff
@@ -3,7 +3,8 @@
 ## in develop
 
 * Command file is now echoed to stderr rather than stdout
-
+* Updates build environment to SBT 1.6.1
```
r-i-v-a, line 6: I don't think we generally put changes that only affect the dev environment into the changelog. @YuxinShi0423 @jdidion

r-i-v-a (changes requested):
I guess the new code style settings are the defaults and I'm ok with it; what does @jdidion think?

No changelog needed for this type of change?

RELEASE_NOTES.md
r-i-v-a: Also, whitespace between end of list & next heading

jdidion:
The new rules are not the default. We should not be arbitrarily changing style formatting rules. If there is a specific rule that someone wants to change that should be handled in a separate PR with substantial justification.

.scalafmt.conf
```diff
@@ -1,7 +1,13 @@
...
+runner.dialect = scala213
 continuationIndent.callSite = 4
 continuationIndent.defnSite = 4
 align.openParenCallSite = true
 align.openParenDefnSite = true
 maxColumn = 100
-assumeStandardLibraryStripMargin = true
\ No newline at end of file
+assumeStandardLibraryStripMargin = true
+danglingParentheses.defnSite = false
```
jdidion: Please remove these rules that you added.

jdidion (changes requested):

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -271,13 +272,12 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
...
         case (name, ExecutableLink(_, _, _, workflow: DxWorkflow)) => (name, workflow.id)
       }
-      .map {
-        case (name, id) =>
-          JsObject(
-              Constants.BundledDependsNameKey -> JsString(name),
-              Constants.BundledDependsIdKey -> JsObject(DxUtils.DxLinkKey -> JsString(id)),
-              Constants.BundledDependsStagesKey -> JsArray(Vector.empty)
-          )
+      .map { case (name, id) =>
```
YuxinShi0423, line 275: we can use `newlines.beforeCurlyLambdaParams = multilineWithCaseOnly` to keep previous format @jdidion

compiler/src/main/scala/dx/compiler/Compiler.scala
```diff
@@ -32,24 +32,39 @@ object Compiler {
...
-  * @param locked whether to generate locked workflows
-  * @param projectWideReuse whether to allow project-wide reuse of applications
-  * @param streamFiles which files to stream vs download
-  * @param waitOnUpload whether to wait for each file upload to complete
-  * @param useManifests whether to use manifest files for all application inputs and outputs
-  * @param complexPathValues whether File and Directory values should be treated as objects
-  * @param fileResolver the FileSourceResolver
-  * @param dxApi the DxApi
-  * @param logger the Logger
+/** Compile IR to native applets and workflows.
```
YuxinShi0423, line 35: we can use `docstrings.style = keep` to keep old format @jdidion

compiler/src/main/scala/dx/compiler/Compiler.scala
```diff
@@ -437,7 +458,8 @@ case class Compiler(extras: Option[Extras],
                                           Some(id),
                                           _,
                                           _,
-                                          _) =>
+                                          _
+                    ) =>
```
YuxinShi0423, line 462:
```
danglingParentheses.defnSite = false
danglingParentheses.ctrlSite = false
danglingParentheses.callSite = false
``` 
can help avoid dangling parentheses like these @jdidion

RELEASE_NOTES.md
YuxinShi0423: Removed. Thanks!

r-i-v-a:
Since there were a lot of code changes since this was opened, I would close it and re-apply the formatting changes on a fresh branch from `develop` (to avoid any mistakes from resolving conflicts). 

For settings / defaults: I am fine with either using the new defaults or making a few changes to approximate our old settings. I'm not opinionated about that, as long as we use the same formatting rules across our repos. @YuxinShi0423

PR #294, opened 2022-02-03
APPS-1053-execute-nested-workflow-step

YuxinShi0423:


commandlinegirl:
It looks like the integration tests are failing @YuxinShi0423

scripts/run_tests.py
```diff
@@ -229,24 +231,12 @@
...
-    "count-lines17-wf",
-    "count-lines15-wf",
-    "count-lines16-wf",
-    "count-lines18-wf",
-]
-
-cwl_conformance_workflows = [ t for t in 
-    [os.path.basename(path)[:-9]
+cwl_conformance_workflows = [
+    os.path.basename(path)[:-9]
```
commandlinegirl: How about changing it to something like: `os.path.basename(path)[:-(len(...))]` so that it's clear what string 9 represents.

compiler/src/test/resources/input_file/bwa-mem-tool_input.json
```diff
@@ -24,6 +24,6 @@
   "bwa-mem-tool.minimum_seed_length": 3,
   "bwa-mem-tool.args.py": {
     "class": "File",
-    "location": "dx://file-G0G3BZQ0yzZf6x1Y3vxzpgk6"
+    "location": "dx://project-Fy9QqgQ0yzZbg9KXKP4Jz6Yq:file-G0G3BZQ0yzZf6x1Y3vxzpgk6"
```
commandlinegirl, line 27: Do the tests pass without being qualified with the project (as before)? Or is this update only an optimization?

commandlinegirl (changes requested):
Only a minor request and one question re file path with projects.

commandlinegirl (dismissed):
LGTM, thank you!

core/src/main/scala/dx/core/languages/cwl/CwlBundle.scala
```diff
@@ -107,7 +107,9 @@ object CwlBundle {
         val newWorkflows = workflows + (wf.name -> wf)
         wf.steps.foldLeft(tools, expressions, newWorkflows, newReqs, newHints, newParents) {
           case ((toolAccu, exprAccu, wfAccu, reqAccu, hintAccu, parentAccu), step) =>
-            val newParent = currentParent ++ CwlDxName.fromDecodedName(step.frag).getDecodedParts
+            val newParent = currentParent ++
```
YuxinShi0423, line 110: https://github.com/dnanexus/dxCompiler/blob/63964655245445f172e360499e53defcd3f57d55/executorCwl/src/main/scala/dx/executor/cwl/CwlWorkflowExecutor.scala#L527-L531

The previous implementation assumed the `step.frag = "nested/step2"` . However if the nested wf has a name ended with 'cwl', it might fail to add the nested wf name to the new parent vector, because now `step.frag=step2` and nested.cwl is recognized as the new namespace where `step2` is defined.

So I add both wf.name and step.name to track the full path from top-level wf to subworkflows here.

executorCwl/src/main/scala/dx/executor/cwl/CwlTaskExecutor.scala
```diff
@@ -373,6 +373,9 @@ case class CwlTaskExecutor(tool: Process,
     updatedSource
   }
 
+  private lazy val rawStep: Option[String] = {
+    targetStep.map(_.split('/').grouped(2).flatMap(_.take(1)).toArray.mkString("/"))
```
YuxinShi0423, line 377: the `targetwf#targetStep` is in the format of `wf#step1/nested/step2`. But `cwltool --single-process` only accept step id sequences like `step1/step2`, so we need to drop the workflow ids.

commandlinegirl (approved):
LGTM, thank you!

PR #300, opened 2022-02-10
APPS-1006 Update include task error trace

Gvaihir:
This PR addresses error propagation from a task to the main process


scripts/run_tests.py
```diff
@@ -36,6 +36,12 @@
     "null-expression2-tool.0",
 }
 
+# these tests are expected to fail at runtime AND throw a specific error message which will be checked
+expected_failure_msg = {
```
commandlinegirl, line 40: That's a useful category, thanks!

commandlinegirl (changes requested):
A minor thing: in `test/subworkflows/` we store tests that are about nested workflows, i.e. workflows calling other workflows in their stages. Nota bene we also have `test/nested/` so we should merge these two at some point. 

I suggest moving your tests, for example to https://github.com/dnanexus/dxCompiler/tree/develop/test/single_tasks

If you haven't done so locally, ie. starting integration tests from your local machine, you can run integration tests from GitHub by adding the `integration` label. Then each new commit will run them, so if you plan more commits you can remove the label and add again at the end. We don't run these tests automatically at commits so as not to generate too much cost.

Thanks!

Gvaihir:
> A minor thing: in `test/subworkflows/` we store tests that are about nested workflows, i.e. workflows calling other workflows in their stages. Nota bene we also have `test/nested/` so we should merge these two at some point.
> 
> I suggest moving your tests, for example to https://github.com/dnanexus/dxCompiler/tree/develop/test/single_tasks

Thanks! Fixed in [7d61a26](https://github.com/dnanexus/dxCompiler/pull/300/commits/7d61a26d0d4059625db3fa15803d1138f48ba279)
 
> If you haven't done so locally, ie. starting integration tests from your local machine, you can run integration tests from GitHub by adding the `integration` label. Then each new commit will run them, so if you plan more commits you can remove the label and add again at the end. We don't run these tests automatically at commits so as not to generate too much cost.

I have triggered integration tests covering 2 new tests only, not the full suite. Will start full integration testing from GHA shortly

commandlinegirl (approved):
LGTM, thank you!

PR #302, opened 2022-02-15
APPS-869 add cwl cromwell test suite

YuxinShi0423:


YuxinShi0423:
Integration test passed using cwl_cromwell test suite:
https://github.com/dnanexus/dxCompiler/runs/5211290741?check_suite_focus=true

scripts/run_tests.py
```diff
@@ -489,36 +489,33 @@
     "cwl_dynamic_initial_workdir",
     "cwl_expressionLib",
     "cwl_format",
-    # "cwl_format_url", # APPS-961 Could not load extension schema https
+    "cwl_format_url", # APPS-1111
```
r-i-v-a: Does the APPS-1111 comment mean this one is still not working?

r-i-v-a (dismissed):
Confused about comment on one test, but it seems the non-ignored tests are all passing. Handing over the review to @commandlinegirl , @Gvaihir

scripts/run_tests.py
```diff
@@ -1545,7 +1542,7 @@ def register_all_tests(verbose: bool) -> None:
         ).endswith("_notimplemented"):
             continue
         for t_file in files:
-            if t_file.endswith(".wdl") or t_file.endswith(".cwl"):
+            if t_file.endswith(".wdl"): # or t_file.endswith(".cwl"):
```
Gvaihir, line 1608: How about a switch case here? Let's say we have a `.blah` workflow language extension in the future and so on - easier to add. Example:
```
extension_switch = {
   ".cwl": do_this_partial_func,
   ".wdl": do_that_partial_func,
   ".json": do_other_partial_func,
}
default = do_nothin_func
exec = extension_switch.get(extension, do_nothin_func)(args_here)
```

Gvaihir (commented):
GG

scripts/run_tests.py
YuxinShi0423: I commented it out because the current compiler only supports the packed version of cwl (as cwl.json). But I like your approach :) maybe we can do it when we refactor the whole run test script?

scripts/run_tests.py
YuxinShi0423: Thanks @r-i-v-a ! I actually fixed the test code but the issue itself in the compiler is not resolved yet. I have reverted my changes to the test and I will bring it back once the bug is fixed.

YuxinShi0423:
All tests added in this ticket as the test suite `cwl_cromwell" passed. @commandlinegirl @Gvaihir

commandlinegirl (approved):
LGTM, we'll need to consider when to run cromwell tests (should they be triggered by a separate label, run at release only or at every PR, etc). I'd say we merge this and create a separate ticket for triggering these and other CWL tests.

Please update the PR and commit message to conform to our convention ("APPS-869 Text text").

Thanks!

Gvaihir (approved):

PR #306, opened 2022-02-24
APPS-1119 Additional fix to optional inputs in fragments

commandlinegirl:


test/bugs/apps_1052_optional_compound_input_wdl10_input.json
```diff
@@ -0,0 +1,3 @@
+{
+  "apps_1052_optional_compound_input_wdl10.t1" : {"inner": {}}
```
YuxinShi0423, line 2: Would that be better if we test it with a valid inner? like t1.inner.bb=111. In that case we are expecting the result a to be 111 instead of 2 and we can check both cases where optional is null and defined. but that is something minor :)

YuxinShi0423 (approved):
Looks great. Thanks! @commandlinegirl

test/bugs/apps_1052_optional_compound_input_wdl10_input.json
commandlinegirl: This is a good point. I've run such a test manually and got the expected result
image
I'll add an automatic test separately.

commandlinegirl: The integration tests passed in the previous commit so I'll go ahead and merge.

PR #313, opened 2022-03-03
APPS-1017 Fix example in ExpertOptions.md

commandlinegirl:
Test: manual, with the fix the example workflow compiles as expected.


Gvaihir (approved):
Thanks!

PR #321, opened 2022-03-15
APPS-1081 fix override docker requirement

YuxinShi0423:
The issue is caused by inconsistent identifier generation in cwljava and cwltool.
In cwljava, processes as a step has id as `#workflowID/step1/processID`, but in cwltool the process id is `#workflowID/step1/run/processID`

It would affect requirement override when using the option `--override` with an override file that specifies the target process ID under the `cwltool:overrides` field.

This PR is to overwrite the requirements in the source cwl instead of using the override file by looking up the process ID.


YuxinShi0423:
TODO: need to fix the JSON lookup: the id of the target task might appear multiple times. Need to use full id instead.

executorCwl/src/main/scala/dx/executor/cwl/CwlTaskExecutor.scala
```diff
@@ -436,11 +436,23 @@ case class CwlTaskExecutor(tool: Process,
...
+    val fullTargetPath = if (targetStep.isDefined) {
+      targetProcess.orElse(root.map(_.name)) match {
+        case Some(targetWf) =>
+          (targetWf +: targetStep.get
+            .split("/")).grouped(2).flatMap(_ :+ s"run").toVector :+ tool.name
+        case None => Vector(tool.name)
+      }
+    } else {
+      Vector(tool.name)
+    }
```
YuxinShi0423, line 449: https://github.com/dnanexus/dxCompiler/blob/1d0c55a4a624e84b35d552c9107b3fe9a61d4475/executorCwl/src/main/scala/dx/executor/cwl/CwlWorkflowExecutor.scala#L527-L532

Gvaihir (dismissed):

Gvaihir (approved):

PR #326, opened 2022-03-28
APPS-1124 restrict applet network access

YuxinShi0423:
Starting from version 1.01 https://github.com/dnanexus/dxCompiler/blob/5f5776005363bb851d833f914f62aab669a23b0c/RELEASE_NOTES.md#L1008-L1011
`dx-docker` is no longer used so full network access should not be given to compiled applets.

- [ ] adding tests with different workflows, e.g. with scatters and differently specified Docker imgs (in the platform, DockerHub, quay,io, etc)


compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -440,24 +440,37 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
...
+            DxAccess.empty.copy(project = Some(DxAccessLevel.Upload),
+                                allProjects = Some(DxAccessLevel.View))
+        )
+      case ExecutableKindWfOutputs(blockPath) =>
+        Some(DxAccess.empty.copy(allProjects = Some(DxAccessLevel.Upload)))
       case _ =>
         // Scatters need network access, because they spawn subjobs that (may) use dx-docker.
         // We end up allowing all applets to use the network
-        Some(DxAccess.empty.copy(network = Some(Vector("*"))))
+        Some(DxAccess.empty)
```
YuxinShi0423, line 453: I added two cases for common input applet and output applet to make sure that 

- the input applet can clone files from other accessible projects
https://documentation.dnanexus.com/developer/api/data-containers/cloning#errors
> VIEW access is required in the source container
> UPLOAD access is required in the destination container
- the output applet can upload to all accessible projects. 

but can be removed from this PR.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -440,24 +440,37 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
...
     // using manifests requires at least UPLOAD access
     val manifestAccess = if (useManifests) {
       Some(DxAccess.empty.copy(project = Some(DxAccessLevel.Upload)))
     } else {
       None
     }
+    val asDeveloperAccess = (applet.kind match {
+      case ExecutableKindNative(_, _, _, _, _) =>
+        Some(DxAccess.empty)
+      case _ => Some(DxAccess.empty.copy(developer = Some(true)))
```
YuxinShi0423, line 464: > If the user launches app1 from the UI or via 'dx run', at that time they are full scoped. The job gets exactly "full scope" ⋂ "dxapp1.json" (where '⋂' means intersection). Then when that job launches app2, the second job gets "full scope" ⋂ "dxapp1.json" ⋂ "dxapp2.json" , etc
And since "full scope" ⋂ "X" is equal to "X", in practice it's the first app that sets the tone
so I added `developer` scope to all non-native executables. But can be removed from this PR.
@commandlinegirl

YuxinShi0423: Also from George:
> It can wrap any DNAnexus native app
And that app may have all sorts of permissions requested in its dxapp.json
The wrapper app(let) MUST have at least that much of permissions in its own dxapp.json, otherwise the system cannot "elevate" permissions from one job to the next
So if the wrapper has lesser permissions, the app that is launched will inherit those stricter permissions

So I wonder if keeping the scope to be `developer` would be something useful...

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
YuxinShi0423: removed

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -441,9 +441,9 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
         // The reorg applet requires higher permissions to organize the output directory.
         Some(DxAccess.empty.copy(project = Some(DxAccessLevel.Contribute)))
       case _ =>
-        // Scatters need network access, because they spawn subjobs that (may) use dx-docker.
-        // We end up allowing all applets to use the network
-        Some(DxAccess.empty.copy(network = Some(Vector("*"))))
+        // ExecutableKindWfFragment / ExecutableKindWfInputs /
+        // ExecutableKindWfOutputs / ExecutableKindWfCustomReorgOutputs
```
YuxinShi0423: There could be cases when file dependences require access to other projects. But since we already state that in the details of the applet
https://github.com/dnanexus/dxCompiler/blob/9c8a47f17c7ede622ef9c2dd0331c9b917b1834c/compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala#L364-L398
we don't have to add access as default here.

Gvaihir:
`Why are your changes focus only on the output job? I thought the network restrictions were asked to be put for all jobs - by default.` - IGNORE THIS
Could you also add tests for the added cases?

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
Gvaihir: pls add tests

Gvaihir (changes requested):

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
YuxinShi0423: Good call. Will do. Thanks!

YuxinShi0423:
Incorrect implementation of "access" requirements: the ones from different sources are not overridden based on precedence but merged. 
https://github.com/dnanexus/dxCompiler/blob/85c60e4ede211441fcb8c93c7103f1dd6f190d92/compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala#L455-L460
https://github.com/dnanexus/dxCompiler/blob/85c60e4ede211441fcb8c93c7103f1dd6f190d92/compiler/src/main/scala/dx/translator/Extras.scala#L195-L197

@commandlinegirl @emiloslavsky I will not fix it in this ticket and the added test cases are based on this current implementation. The issue has been added to ticket https://jira.internal.dnanexus.com/browse/APPS-1148

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -659,6 +657,17 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
         )
       case _ => throw new Exception("Missing runSpec")
     }
+
+    desc.access shouldBe Some(
+        JsObject(
+            Map(
+                "project" -> JsString("ADMINISTER"),
+                "allProjects" -> JsString("VIEW"),
```
YuxinShi0423, line 689: settings from extras.json as default task attributes.

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -711,6 +721,149 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
         )
       case _ => throw new Exception("Missing runSpec")
     }
+    desc.access shouldBe Some(
+        JsObject(
+            Map(
+                "project" -> JsString("VIEW"),
+                "allProjects" -> JsString("VIEW"),
```
YuxinShi0423, line 752: settings from per task attributes.

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -711,6 +721,149 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
+        Set(
+            Field.Access
+        )
+    )
+
+    desc2.access shouldBe Some(
+        JsObject(
+            Map(
+                "allProjects" -> JsString("VIEW"),
+                "network" -> JsArray(Vector(JsString("*")))
```
YuxinShi0423, line 798: two tasks are compiled in this test wdl. The applet1 does not require a docker so the network setting is empty. Applet 2 requires external docker so network is "*".

Both are given all projects "VIEW" permission since "dockerRegistry" is required by extras.json.

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -711,6 +721,149 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
+    taskDesc.access shouldBe Some(
+        JsObject(
+            Map(
+                "allProjects" -> JsString("VIEW"),
+                "network" -> JsArray(Vector(JsString("*")))
+            )
+        )
+    )
+
+    val dxNativeApp = dxApi.resolveApp(
```
YuxinShi0423: In this workflow we have 3 kinds of executables:

- frag applet: `network = []`
- task applet called by frag applet: `network = ["*"]` since it requires external docker
- native app: no change is applied to the native app

YuxinShi0423:
@Gvaihir @commandlinegirl I added some test cases. Feel free to comment :D

compiler/src/test/scala/dx/compiler/CompilerTest.scala
commandlinegirl: Sounds good

commandlinegirl (approved):
LGTM, thank you!

compiler/src/main/scala/dx/translator/cwl/ProcessTranslator.scala
```diff
@@ -575,14 +575,24 @@ case class ProcessTranslator(cwlBundle: CwlBundle,
...
-          standAloneWorkflow
+          name = s"${wfName}_frag_${getStageId()}",
+          inputs = inputParams,
+          outputs = outputParams,
+          instanceType = instanceType,
+          container = container,
+          kind = ExecutableKindWfFragment(calleeName, blockPath, fqnDictTypes, scatterChunkSize),
+          document = standAloneWorkflow,
+          attributes = attributes,
+          requirements = requirements
```
YuxinShi0423, line 595: In the previous integration tests two wdl tests could not pass, because the embedded task applet requires external network access but the frag does not have it.

I checked the implementation of applet/new https://github.com/dnanexus/nucleus/blob/db5a6120c97957c3c1a7c1c3d9625c4d32ea506c/apiserver/execution/applet._js#L32-L46
By default `network=[]`, so the frag will have None network access, while the inner one was given it during compilation
However, when running the applet, the network access is inherited from parent job:
https://github.com/dnanexus/nucleus/blob/dc128eed411a320ffae04952655008162c9307ac/apiserver/execution/exec_handlers/applets._js#L191-L204

So here I pulled the info from the callee applet and copied them to the frag applet.

YuxinShi0423: @commandlinegirl @Gvaihir

compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala
```diff
@@ -681,14 +681,25 @@ case class CallableTranslator(wdlBundle: WdlBundle,
...
-          standAloneFrag
+          name = s"${wfName}_frag_${getStageId()}",
+          inputs = inputParams,
+          outputs = outputParams,
+          instanceType = instanceType,
+          container = container,
+          kind = ExecutableKindWfFragment(innerCall, blockPath, fqnDictTypes, scatterChunkSize),
+          document = standAloneFrag,
+          attributes = attributes,
+          requirements = requirements
```
YuxinShi0423, line 702: same as above, just for cwl

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -235,34 +235,38 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
       case _ => Map.empty
     }
     // Add platform Docker image dependency to bundledDepends
-    val bundledDependsDocker: Option[JsValue] = applet.container match {
-      case DxFileDockerImage(_, dxfile) => {
-        val id = dxfile.id
+    val bundledDependsDocker: Option[JsValue] = if (applet.kind == ExecutableKindApplet) {
```
YuxinShi0423, line 238: Since both frag applet and the task applet have the container info, exclude the frag applet when making `bundledDepends`
@commandlinegirl @Gvaihir

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -305,11 +309,15 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
       case _ => Map.empty
     }
     // Add Docker image info to details
-    val dockerImageDetails: Map[String, JsValue] = applet.container match {
-      case DxFileDockerImage(_, dxfile) => Map(Constants.DockerImage -> dxfile.asJson)
-      case NetworkDockerImage(pullName) => Map(Constants.NetworkDockerImage -> JsString(pullName))
-      case DynamicDockerImage           => Map(Constants.DynamicDockerImage -> JsBoolean(true))
-      case NoImage                      => Map.empty
+    val dockerImageDetails: Map[String, JsValue] = if (applet.kind == ExecutableKindApplet) {
```
YuxinShi0423, line 312: same here. Only for task applet but not frag applet.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -430,7 +438,7 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
     }
     // update depending on applet type
     val appletKindAccess = applet.kind match {
-      case ExecutableKindApplet =>
+      case ExecutableKindApplet | ExecutableKindWfFragment(_, _, _, _) =>
```
YuxinShi0423, line 441: Give both the task applet and the wrap frag applet network access.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
commandlinegirl: Not sure I 100% follow, where the frag applet will get its `bundledDepends` info from?

YuxinShi0423: https://github.com/dnanexus/dxCompiler/blob/3a8f4cd17f94356d8b4960317ecfea158fefd45f/compiler/src/main/scala/dx/translator/cwl/ProcessTranslator.scala#L578-L596

Here `availableDependencies` already contains the embedded applets (we translate them first then make the wrapper frag applet when translating the wf)

So here we use the container info fetched from the inner applet to make the new frag applet.

commandlinegirl (approved):

commandlinegirl (approved):

commandlinegirl:
(I approved twice since there was a glitch in GitHub and I got 500 after the first one)

commandlinegirl (approved):

compiler/src/main/scala/dx/translator/cwl/ProcessTranslator.scala
```diff
@@ -575,14 +575,24 @@ case class ProcessTranslator(cwlBundle: CwlBundle,
         param.name -> param.dxType
       }.toMap
 
+      val (instanceType, container, attributes, requirements) = availableDependencies
```
Gvaihir, line 578: Please consider the following:
1. Creating an abstract class for `ProcessTranslator` and `CallableTranslator`.
2. Creating a common method for the line above. 
3. Not return tuples. They are hard to figure out what's happening and the return types.

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
Gvaihir: Does this create any inefficiencies? Why do we want to create additional dependencies between the components?

compiler/src/main/scala/dx/translator/cwl/ProcessTranslator.scala
Gvaihir: Isn't this component responsible for CWL? How is this connected to WDL integration test?

Gvaihir (commented):
Since @commandlinegirl already approved - would be good in the future to avoid code repetition wherever one can create parent class with common methods

commandlinegirl:
> Since @commandlinegirl already approved - would be good in the future to avoid code repetition wherever one can create parent class with common methods

Thanks @Gvaihir, this could be a scope of a separate, refactoring task.

Gvaihir:
> > Since @commandlinegirl already approved - would be good in the future to avoid code repetition wherever one can create parent class with common methods
> 
> Thanks @Gvaihir, this could be a scope of a separate, refactoring task.

Heh? Not sure why my review is a request for changes. I just left it as a comment..
EDIT: those were my previous requests, apparently

Gvaihir (approved):

compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -441,9 +449,8 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
         // The reorg applet requires higher permissions to organize the output directory.
         Some(DxAccess.empty.copy(project = Some(DxAccessLevel.Contribute)))
       case _ =>
-        // Scatters need network access, because they spawn subjobs that (may) use dx-docker.
-        // We end up allowing all applets to use the network
-        Some(DxAccess.empty.copy(network = Some(Vector("*"))))
```
YuxinShi0423, line 446: @commandlinegirl here we can just make a new case to give `ExecutableKindWfFragment` full access and restrict the other kinds  `ExecutableKindWfInputs / ExecutableKindWfOutputs / ExecutableKindWfCustomReorgOutputs`

PR #329, opened 2022-04-05
APPS-1169 fix ignore reuse precedence

YuxinShi0423:


compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala
```diff
@@ -652,12 +652,13 @@ case class ApplicationCompiler(typeAliases: Map[String, Type],
...
-        case IgnoreReuseRequirement(value) => value
-      }
+    val ignoreReuse = extras
+      .flatMap(_.ignoreReuse)
       .orElse(
-          extras.flatMap(_.ignoreReuse)
+          applet.requirements
+            .collectFirst {
+              case IgnoreReuseRequirement(value) => value
+            }
```
YuxinShi0423, line 661: what's provided in extras should override what is given in the task runtime.

The workflowCompiler does not have this issue, because it does not have its own runtime section therefore the only ignoreReuse comes from extras.json

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -625,7 +625,31 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
     desc.ignoreReuse shouldBe Some(true)
   }
 
-  it should "be able to include runtime hints and override extras global" in {
+  it should "timeout can be overriden from the extras file" taggedAs NativeTest in {
```
YuxinShi0423, line 628: move to here so all tests related to timeout setting are put together.

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -1119,8 +1119,32 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
     ignoreReuseFlag shouldBe Some(JsBoolean(true))
   }
 
+  it should "override job-reuse flag on applet" taggedAs NativeTest in {
```
YuxinShi0423, line 1122: add test on ApplicationCompiler to cover the issue: the `ignoreReuse=false` from extras should override the task runtime setting.
https://github.com/dnanexus/dxCompiler/blob/85c60e4ede211441fcb8c93c7103f1dd6f190d92/compiler/src/test/resources/compiler/add_runtime_hints.wdl#L23

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -1119,8 +1119,32 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
+    // make sure the job reuse flag is set
+    val (_, stdout, _) =
+      SysUtils.execCommand(s"dx describe ${dxTestProject.id}:${appletId} --json")
+    val ignoreReuseFlag = stdout.parseJson.asJsObject.fields.get("ignoreReuse")
+    ignoreReuseFlag shouldBe Some(JsBoolean(false))
+  }
+
   it should "set job-reuse flag on workflow" taggedAs NativeTest in {
-    val path = pathFromBasename("subworkflows", basename = "trains_station.wdl")
+    val path = pathFromBasename("subworkflows", basename = "check_route.wdl")
```
YuxinShi0423, line 1147: use this wdl instead which makes it easier to check if the flag is passed to embedded tasks.

test/special_flags/add2_extras.json
```diff
@@ -1,3 +1,4 @@
 {
+  "ignoreReuse": true,
```
YuxinShi0423, line 2: this flag used to be here but was removed, but `add2` is supposed to be the test for `ignoreReuse`
https://github.com/dnanexus/dxCompiler/blob/85c60e4ede211441fcb8c93c7103f1dd6f190d92/scripts/run_tests.py#L263-L267

@Gvaihir I am thinking about checking the job description to see if these flags are working. But would that be something you already suggested in refactoring the integration tests?

Gvaihir: I haven implemented any specific class of tests. They all will be subclasses of one parent class, extendable to any scenario. Main goal - whatever test subclass you add it will be easy to test that your addition works. So for now go as you plan

Gvaihir (dismissed):
gg

commandlinegirl (approved):
LGTM, thanks!

test/cwl_cromwell/cwls/cwl_scatter-wf1.cwl.json
```diff
@@ -60,7 +61,7 @@
                             "loadContents": true,
                             "outputEval": "$(self[0].contents)"
                         },
-                        "id": "#main/step1/run/echo_out"
+                        "id": "echo_out"
```
commandlinegirl, line 64: One questions: why do we need to update these paths (ids)?

PR #333, opened 2022-04-18
APPS 1128 Update native app instance override

Gvaihir:
Overriding instance types for native app(let)s if they are wrapped in fragments (conditionals and scatters)


doc/ExpertOptions.md
```diff
@@ -1067,6 +1067,20 @@ task concat {
   }
 }
 ```
+### Unsupported overrides
+Current version of dxCompiler **does not** support instance override for native app(let)s in the following cases:
+1. Specifications of system requirements (RAM/CPU/storage/etc.) is not supported. Use instance names instead e.g. `mem1_ssd1_v2_x4` (instance names may be not transferable between regions)
+2. Instance override by string interpolation: e.g. `mem1_ssd1_~{REMAINING_PART_OF_INSTANCE_NAME}`
```
commandlinegirl: What would be an example of the value of REMAINING_PART_OF_INSTANCE_NAME? You mean, for example `mem1_ssd1_~x2`?

executorWdl/src/test/scala/dx/executor/wdl/WdlTaskExecutorTest.scala
```diff
@@ -255,13 +255,13 @@ class WdlTaskExecutorTest extends AnyFlatSpec with Matchers {
...
-    // update paths of input files - this requires a round-trip de/ser
+    // update paths of input files - this requires a round-trip
     // JSON -> IR -> WDL -> update paths -> IR -> JSON
-    // which requires some auxilliarly objects
+    // which requires some auxiliary objects
     val outputSerializer: ParameterLinkSerializer = ParameterLinkSerializer(fileResolver, dxApi)
     val taskInputs: Map[DxName, TAT.InputParameter] =
       task.inputs.map(inp => WdlDxName.fromDecodedName(inp.name) -> inp).toMap
-    val updatedInputs = inputDeserializer
+    val updatedInputs: Map[DxName, JsValue] = inputDeserializer
```
commandlinegirl, line 264: Thanks for improving readability!

test/nested/apps_1128_frag_native_instance_type_override.wdl
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  command <<< >>>
+
+  output {
+    File file = "placeholder.txt"
+  }
+
+  runtime {
+    dx_instance_type: "mem1_ssd1_v2_x16"
```
commandlinegirl: If the intention is to run this workflow, let's use a cheaper instance type, e.g. mem1_ssd1_v2_x2 (since the default for the concatonator is mem2_hdd2_x4)

test/nested/apps_1128_frag_native_instance_type_override.wdl
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  command <<< >>>
+
+  output {
+    File file = "placeholder.txt"
+  }
+
+  runtime {
+    dx_instance_type: "mem1_ssd1_v2_x8"
```
commandlinegirl: How about mem1_ssd1_v2_x4 here?

test/nested/apps_1128_frag_native_instance_type_override.wdl
```diff
@@ -0,0 +1,113 @@
+version 1.1
```
commandlinegirl: It doesn't look like this test is used (?). You'd need to verify the job description that the instance type was overridden with the expected value.

test/nested/apps_1128_frag_native_instance_type_override.wdl
```diff
@@ -0,0 +1,113 @@
...
+
+workflow override {
```
commandlinegirl: Our convention is to use the name of the workflow that matches the name of the file (here `apps_1128_frag_native_instance_type_override`) so that it's easy to find the analysis in the test dxCompiler_playground project.

test/nested/apps_1128_frag_native_instance_type_override.wdl
```diff
@@ -0,0 +1,113 @@
...
+  command <<<
+    echo "a" > input_a.txt
+    echo "b" > input_b.txt
+  >>>
+  output {
+    Array[File] inputs = ["input_a.txt", "input_b.txt"]
+  }
+}
+
+task default {
```
commandlinegirl: It is likely some other workflow has a task with the same name, which will result in a conflict when running integration tests (the applet of a workflow compiled first will be overridden by the applet with the same name from a workflow compiled afterwards), so it's better to use more specific names.

RELEASE_NOTES.md
```diff
@@ -8,6 +8,7 @@
...
 * Updates to the documentation, esp. sections about [delay workspace destruction](https://github.com/dnanexus/dxCompiler/blob/444036acb16f2555d3cfe5f4c892b9996a8079dc/doc/ExpertOptions.md#delay-workspace-destruction), [DNAnexus-specific runtime settings](https://github.com/dnanexus/dxCompiler/blob/444036acb16f2555d3cfe5f4c892b9996a8079dc/doc/ExpertOptions.md#additional-dnanexus-specific-runtime-settings), and [outputing DNAnexus files](https://github.com/dnanexus/dxCompiler/blob/444036acb16f2555d3cfe5f4c892b9996a8079dc/doc/ExpertOptions.md#dnanexus-files-as-outputs).
 * WDL: Fix to configuring network access to tasks (applets). Now you can disable network access to your tasks with the `runSpec.access` setting in the `extras.json` file:
+* WDL: Fix to native app(let)s instance override. Now native app(let)s wrapped in scatters and conditionals get their default instances overridden with static dnanexus names. Dynamic instance override is not yet supported. Override with compute requirements (CPU/RAM/etc.) is not supported.
```
commandlinegirl: I'm cutting a release now, so if you could merge this (and update release notes) after https://github.com/dnanexus/dxCompiler/pull/338 is merged?

doc/ExpertOptions.md
```diff
@@ -1067,6 +1067,20 @@ task concat {
   }
 }
 ```
+### Unsupported overrides
+Current version of dxCompiler **does not** support instance override for native app(let)s in the following cases:
+1. Specifications of system requirements (RAM/CPU/storage/etc.) is not supported. Use instance names instead e.g. `mem1_ssd1_v2_x4` (instance names may be not transferable between regions)
```
commandlinegirl: IMHO, we can remove "(instance names may be not transferable between regions)", or add it in the section above, where we first refer to the naming. Otherwise it may be confusing why we mention regions here.

commandlinegirl (changes requested):
Most are minor comments, thanks!

doc/ExpertOptions.md
Gvaihir: Yep, the remaining part of the instance name, like `v2_x4` in `mem1_ssd1_v2_x4`. So in the task there will be something like 
```
Int REMAINING_PART_OF_INSTANCE_NAME = v2_x4
runtime {
...
dx_instance_type: "mem1_ssd1_~{REMAINING_PART_OF_INSTANCE_NAME}"
}
```

Added more explanation in [a8c3b98](https://github.com/dnanexus/dxCompiler/pull/333/commits/a8c3b98ca341d3808aaa2e932ef970e79f985abe)

test/nested/apps_1128_frag_native_instance_type_override.wdl
Gvaihir: fixed in [2cb5022](https://github.com/dnanexus/dxCompiler/pull/333/commits/2cb5022b51c0c6564aadb4bed60becb77f5fe37c)

test/nested/apps_1128_frag_native_instance_type_override.wdl
Gvaihir: fixed in [2cb5022](https://github.com/dnanexus/dxCompiler/pull/333/commits/2cb5022b51c0c6564aadb4bed60becb77f5fe37c)

test/nested/apps_1128_frag_native_instance_type_override.wdl
Gvaihir: fixed in [954b716](https://github.com/dnanexus/dxCompiler/pull/333/commits/954b716d16a2e701d572de97e45e206a498eb83d)

test/nested/apps_1128_frag_native_instance_type_override.wdl
Gvaihir: fixed in [15cc191](https://github.com/dnanexus/dxCompiler/pull/333/commits/15cc191bb8ec1b4c6fed9c4abe0963b3291562ce)

RELEASE_NOTES.md
Gvaihir: fixed in [15cc191](https://github.com/dnanexus/dxCompiler/pull/333/commits/15cc191bb8ec1b4c6fed9c4abe0963b3291562ce)

doc/ExpertOptions.md
Gvaihir: fixed in [15cc191](https://github.com/dnanexus/dxCompiler/pull/333/commits/15cc191bb8ec1b4c6fed9c4abe0963b3291562ce), added reference to Global Workflow Recommendations

test/nested/apps_1128_frag_native_instance_type_override.wdl
Gvaihir: Current integration tests don't have this functionality. [This WF](https://github.com/dnanexus/dxCompiler/blob/69cb26ce9a2e67aedc91bbc0442e52da7ad5de52/test/wdl_1_0/instance_types.wdl) exploits a workaround to output instance properties with stdout of the task command. I'll have to use a different native app to do something similar (because you can't really change outputs of a task with the native app). I was planning to add this regression as a part of the new suite later and that would automate this test.
However, having no regression bothers me. I will create an applet that outputs instance type and use that in the workflow - better keep this regression in place, because there will be more work on instance overrides.

Gvaihir: included regression. Applet that's used is in the `file-apps` PR
[0db9df3](https://github.com/dnanexus/dxCompiler/pull/333/commits/0db9df3add14a1baa6ba2ddcc7760a9514e25be8)

commandlinegirl (dismissed):
LGTM, thanks!

test/nested/apps_1128_frag_native_instance_type_override.wdl
commandlinegirl: Thanks!

Gvaihir:
Added a few things and a regression test. Running integration test. @commandlinegirl sorry that disregarded your approval

commandlinegirl (approved):

PR #336, opened 2022-04-22
APPS 1177 Update native instance override via systemRequirements

Gvaihir:
Investigation how to use sys requirements to override native app instance AND keep default instance if no overriding.


commandlinegirl:
Added a comment to the corresponding JIRA ticket.

PR #339, opened 2022-04-26
APPS-1177 Update native instance override via system requirements

Gvaihir:
Had to close previous PR for this ticket due to conflicts.


compiler/src/test/resources/bugs/apps_1177_native_indirect_override_unit.wdl
```diff
@@ -0,0 +1,75 @@
...
+
+    output {
+        String mock_out = ""
+    }
+
+    runtime {
+        dx_app: object {
+                    type: "app",
+                    id: "app-G6G0jX80g1FZX1Z57z3zbG6v",
+                    name: "mock_app_sciprodbuild/0.0.6"
```
commandlinegirl: This app is not accessible to me so perhaps it could be replaced with a public app on staging env, so it's available to whoever runs this (if they want to describe the app or use this script in integration tests as well).

test/nested/apps_1177_native_indirect_override.wdl
```diff
@@ -0,0 +1,114 @@
+version 1.1
+
+workflow apps_1177_native_indirect_override {
+
+  call apps_1177_default_instance {
+    input:
+      mock_input = "apps_1177_default_isntance"
```
commandlinegirl: Minor: typo isntance

test/nested/apps_1177_native_indirect_override.wdl
```diff
@@ -0,0 +1,114 @@
...
+
+  output {
+    String instance_name = ""
+  }
+
+  runtime {
+    memory: "16 GB"
+    dx_app: object {
+              type: "applet",
+              id: "applet-G9VZBF00yzZgj97K42BQKJ7Z",
```
commandlinegirl: How about creating a folder in `dxCompiler_playground` with applets used in tests? It can be under the existing `test_data` or a sibling to it, e.g. `test_executables` or something similar?

commandlinegirl (dismissed):
LGTM, thank you!

test/nested/apps_1177_native_indirect_override.wdl
Gvaihir: moved to `/test_executables`, although it's not going be used when the test suite is ready - all assertions of runSpec will be happening locally

compiler/src/test/resources/bugs/apps_1177_native_indirect_override_unit.wdl
Gvaihir: I'll try to add apps team to the allowed users. It just seems as a very handy lightweight app that needs no inputs and runs nothing.

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -557,31 +560,31 @@ case class WorkflowCompiler(separateOutputs: Boolean,
...
+          case app: Application =>
+            val instanceType: String = app.instanceType match {
+              case _: StaticInstanceType
+                  if (app.kind.getClass == classOf[ExecutableKindWfFragment]) =>
+                "do not override"
+              case static: StaticInstanceType => instanceTypeDb.apply(static.req).name
+              case DefaultInstanceType if (app.kind.getClass == classOf[ExecutableKindNative]) =>
+                "do not override"
+              case _ => instanceTypeDb.defaultInstanceType.name
+            }
```
YuxinShi0423, line 575: Not sure if it should be considered, but in `ApplicationCompiler`
https://github.com/dnanexus/dxCompiler/blob/02a8ed68b27a187c134dded33334ee756c4ceeac/compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala#L178-L187
`defaultInstanceType` is prior to the `instanceTypeDb.defaultInstanceType`. Shall we also assign the stage default here in the same way?

core/src/main/scala/dx/core/languages/wdl/Runtime.scala
```diff
@@ -122,12 +122,26 @@ case class Runtime(wdlVersion: WdlVersion,
...
+        val usedGpuDefault: Boolean = gpu match {
+          case Some(b) => false
+          case None    => true
+        }
+        val usedOtherSysDefaults: Boolean = {
+          runtimeAttrs.runtime.map(_.isDefaultSystemRequirements) match {
+            case Some(b) => b
+            case None    => false
+          }
+        }
```
YuxinShi0423, line 134: Just one trivial case where we do have the cpu/disks/memory specified but they happen to be equal to the default. Could that happen?

compiler/src/test/resources/bugs/apps_1177_native_indirect_override_unit.wdl
Gvaihir: fixed in [6e71262](https://github.com/dnanexus/dxCompiler/pull/339/commits/6e712625cf217365e57d565b513ed83d6731457f), using applet APPS_1177

test/nested/apps_1177_native_indirect_override.wdl
Gvaihir: fixed in [6e71262](https://github.com/dnanexus/dxCompiler/pull/339/commits/6e712625cf217365e57d565b513ed83d6731457f)

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
Gvaihir: In `ApplicationCompiler` the run spec is created, here it's getting overridden by whatever custom instance is requested. Since native app is not compiled by `ApplicationCompiler` - its instance override has to happen later. In this case it was implemented (earlier by someone) for the new instance request to be specified in the stage's `systemRequirements` attr, which presumably has a precedence over `instanceType` attr when the stage is compiled. However, if no overrides were in the WF script, we need to roll back to the native exec's defaults. It's done by NOT specifying `systemRequirements`. I chose to resolve it to None similar to the [`case _` here](https://github.com/dnanexus/dxCompiler/blob/7054ab5b1cdbb90a7162e48dc40b53c18f2054e1/compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala#L588) (for the defaults of the native execs) by this arbitrary string to avoid implementation of something like `dxApi.appdescribe()` -> then parsing a non-trivial nested json that is an app description. But it could be an alternative.

core/src/main/scala/dx/core/languages/wdl/Runtime.scala
Gvaihir: Yes, thanks, in this case it will create an `InstanceTypeRequest` with those requirements and resolve to the instance type that satisfy those requirements. But that will be an explicit override coming form a WF script. This is to resolve to native exec's defaults if there are no overrides

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
Gvaihir: I hope that answered your question. If you were asking just to have same case statements, that's gonna be difficult because these 2 situations try to resolve different cases, and may have different spectrums of inputs.

core/src/test/scala/dx/core/languages/wdl/InstanceTypesTest.scala
```diff
@@ -304,6 +299,10 @@ class InstanceTypesTest extends AnyFlatSpec with Matchers {
                           os = Some(Constants.DefaultExecutionEnvironment))
   }
 
+  it should "create a runtime with default instance type" in {
```
Gvaihir, line 302: Created a separate test case for convenience. 
Also resolve to empty `InstanceTypeRequest` if no instance override is happening, which is ultimately resolved to `InstanceTypeDB.default`

commandlinegirl (approved):

Issue #182, opened 2021-09-09
sub-workflow defaults not overridden by provided value


Ben-Habermeyer:
Hello, I noticed in testing dxCompiler 2.4.10 that sub-workflow default input values are not overridden by a different provided value. 

For example with the following 2 WDL files

**workflow1.wdl**
```
import "workflow2.wdl"

workflow workflow1 {
    input {
    ...
    }
   call workflow2 {
       input:
            my_input=["hello"]
   }
...
}
```

**workflow2.wdl**
```
workflow workflow2 {
    input {
        Array[String] my_input = []
    }
...
}
```
 
when workflow2 is run, it is using `[]` rather than `["hello"]`

emiloslavsky:
For reference: https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#task-inputs:
```
A task's input section declares its input parameters. The values for declarations within the input section may be specified by the caller of the task. An input declaration may be initialized to a default value or expression that will be used when the caller does not specify a value
```
so when workflow1 invokes workflow2, workflow2 should execute with `my_input = ["hello"]`.
@commandlinegirl let's see if 2.10.1 has the correct behavior

commandlinegirl:
I tested this with dxCompiler 2.10.1 and observe the correct behavior (I extended our test for subworkflow defaults with arrays, to be sure, [here](https://github.com/dnanexus/dxCompiler/pull/341/files). If you see an issue again, please ping us, thank you!

commandlinegirl:
Hmm, I just tested it additionally with dxCompiler 2.4.10 (using this updated [test](https://github.com/dnanexus/dxCompiler/pull/341/files)) and cannot reproduce this issue. I'll try to print the values inside the subworkflow instead of only passing them to the outputs.

commandlinegirl:
I verified that the issue was fixed in [dxCompiler-2.7.0](https://github.com/dnanexus/dxCompiler/releases/tag/2.7.0) and updated the release notes for that release. The input values before were indeed not used inside the subworkflow, but starting from 2.7.0 the behavior is correct. cc @emiloslavsky. Our integration [test](https://github.com/dnanexus/dxCompiler/pull/341/files) is updated to cover that.

Ben-Habermeyer:
great thanks @commandlinegirl and @emiloslavsky !

PR #341, opened 2022-05-04
APPS-1167 Default array values of subworkflow should be overridden by provided value (test added)

commandlinegirl:
This test now verified that the issue https://github.com/dnanexus/dxCompiler/issues/182 has been fixed.


Gvaihir (dismissed):
All good, thanks! Maybe makes sense to rename some tasks to make applets/WFs with unique names?

commandlinegirl:
> All good, thanks! Maybe makes sense to rename some tasks to make applets/WFs with unique names?

Done. Also added a task that verifies the provided values are indeed used in the subworkflow, as mentioned here https://github.com/dnanexus/dxCompiler/issues/182

Gvaihir (approved):

PR #342, opened 2022-05-10
APPS-1175 Update uploading intermediate outputs for nested workflows

Gvaihir:
WIP:
Default destination for the file uploads is in the execution container, not the project. Trying to figure out the right way to fix, options may be:
1. Make a frag applet aware of the intermediate outputs and push all the outputs to the container of the parent execution. This looks like the most viable option requiring the least modifications. 

2. For the non-top level workflows called by a frag exec - register intermediate outputs as workflow outputs.

3. For the non-top level workflows called by a frag exec - create a collector applet that registers all intermediate outputs and pushes them to the top workflow

 

Options 2 and 3 are possible if we introduce a way to distinguish top level vs lower level workflows. So far all nested workflows (except frags) are top level. Even if a workflow is at the lower level 
UPDATE: This comment is outdated. I chose option 2 for implementation, bypassing the notion that even nested WFs are top level from the perspective of the dxCompiler.


Gvaihir:
There will be changes to the code after APPS-1217 is approved. Will be adding conditionals and feature flags.

Gvaihir:
Integration test `complex_manifest` due to current incompatibility of manifests with output propagation from nested WF. This depends on APPS-1217

Gvaihir:
> Integration test `complex_manifest` due to current incompatibility of manifests with output propagation from nested WF. This depends on APPS-1217

fixed in [040356d](https://github.com/dnanexus/dxCompiler/pull/342/commits/040356d1c5f93da2f1b606fbb1d00a4c682f50f7)

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -80,7 +80,7 @@ case class WorkflowCompiler(separateOutputs: Boolean,
           ParameterLinkStage(dxStage, IORef.Output, param.name, dxType, param.dxType)
         case StageInputWorkflowLink(wfParam) =>
           // TODO: if the input has a non-static default, link to the value of the workflow input
-          //  (either the user-specified value or the result of evaluting the expression)
+          //  (either the user-specified value or the result of evaluating the expression)
```
commandlinegirl, line 83: Thanks for fixing the typos as well!

executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala
```diff
@@ -528,7 +528,11 @@ case class WdlWorkflowExecutor(docSource: FileNode,
         launchCall(callInputs, Some(blockIndex.toString))
       // TODO: right now, we export all outputs of the executable, which may include intermediate
       //  outputs. Should we filter executable outputs using call.wdlType.output?
-      jobMeta.createExecutionOutputLinks(dxExecution, executableLink.outputs, Some(callName))
+      jobMeta.createExecutionOutputLinks(
```
commandlinegirl, line 531: Thanks!

test/nested/apps_1175_outs_not_uploaded_outer_results.json
```diff
@@ -0,0 +1,4 @@
+{
+  "apps_1175_outs_not_uploaded_outer.stage-1.ni___test_inner1_stage_0_integration_check": "It works: upload_intermediateupload_intermediate",
+  "apps_1175_outs_not_uploaded_outer.apps_1175_outs_not_uploaded_outer_frag_stage_0_stage_1_ni___test_inner1_stage_0_integration_check": "It works: upload_intermediateupload_intermediate"
```
commandlinegirl, line 3: Have the output names always been so long (such multiple IDs concatenation)?

commandlinegirl (commented):
This is a partial review

test/nested/apps_1175_outs_not_uploaded_outer_results.json
Gvaihir: Nope, this is introduced with the new functionality, but only for intermediate outputs. The reason for that is to make them unique so that there are no conflicts if a name repeats across multiple namespaces (e.g. task t1 can have `results` output and task t2 can have `results`). If there are duplicates, there will be an applet compilation error (platform behavior). So to differentiate those outputs, I'm prepending a namespace that includes stage name and stage id.

commandlinegirl:
> Options 2 and 3 are possible if we introduce a way to distinguish top level vs lower level workflows. So far all nested workflows (except frags) are top level. Even if a workflow is at the lower level

Not sure I understand the last two sentences

core/src/main/scala/dx/core/languages/wdl/WdlCodeGenerator.scala
```diff
@@ -194,7 +201,12 @@ case class CodeGenerator(typeAliases: Map[String, WdlTypes.T_Struct],
         .map { parameter =>
           val wdlType = WdlUtils.fromIRType(parameter.dxType, typeAliases)
           val defaultVal = WdlUtils.getDefaultValueOfType(wdlType)
-          TAT.OutputParameter(parameter.name.decoded, wdlType, defaultVal)(SourceLocation.empty)
+          TAT.OutputParameter(
+              DxName.disallowedCharsRegex
+                .replaceAllIn(parameter.name.decoded, "_"), // Hack to ensure proper output names in APPS-1175. Can't use .encoded because the subsequent translation will fail
```
commandlinegirl: IMHO, the comment suggests it's only applied for the APPS-1175 test but this is used to for qualification of all intermediate outputs (?)

compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala
```diff
@@ -326,6 +353,129 @@ case class CallableTranslator(wdlBundle: WdlBundle,
...
+      * also calls that are executed under the alias names if the original names are among dependencyOutputs, like:
+      * `call abc as xyz {}`
+      * because if we don't, those outputs will not be recognized/propagated from lower levels of the task/wf executions
+      * leading to the platform I/O errors at the runtime. This function does those eliminations. The search and
+      * recursion is performed as A.B.C -> lookup (A.B) + C.
+      * @param name: output parameter name
+      * @param aliases: Block.blockNames. unqualifiedName -> actualName
+      * @param substitutedName: Name accumulator, using WdlDxName because it's more restrictive. Seed with a mock name
+      *                         which will be popped upon exit. Hack because DxName does not allow empty names.
+      * TODO: add Cwl functionality.
```
commandlinegirl: Please remove this and create a corresponding JIRA ticket. Thanks!

test/nested/apps_1175_outs_not_uploaded_outer_results.json
commandlinegirl: Got it, makes sense.

commandlinegirl:
> WIP: Default destination for the file uploads is in the execution container, not the project. Trying to figure out the right way to fix, options may be:
> 
> 1. Make a frag applet aware of the intermediate outputs and push all the outputs to the container of the parent execution. This looks like the most viable option requiring the least modifications.
> 2. For the non-top level workflows called by a frag exec - register intermediate outputs as workflow outputs.
> 3. For the non-top level workflows called by a frag exec - create a collector applet that registers all intermediate outputs and pushes them to the top workflow

Would it be possible to list in the description in what scenarios the propagation to the top-level (parent) workflow is needed (and being added), e.g. when a workflow is run from an applet but not when a regular sub-workflow is called in a stage.

commandlinegirl (dismissed):
LGTM modulo a few minor requests

Gvaihir:
> > Options 2 and 3 are possible if we introduce a way to distinguish top level vs lower level workflows. So far all nested workflows (except frags) are top level. Even if a workflow is at the lower level
> 
> Not sure I understand the last two sentences

Opening comment was outdated, I made the updates. Also please check the comments to the APPS-1175, I elaborate there to a better extent. 
Basically what I say in the part above: nested workflows are still considered top level workflows by dxCompiler. Meaning they are translated by [`WdlWorkflowTranslator.translateTopWorkflowLocked()](https://github.com/dnanexus/dxCompiler/blob/c7b3061e49f5e7f413e0481fd880a8393655a256/compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala#L1081) despite being called from a parent workflow. 
Regardless of that, my implementation does not really care whether it's a top level or not. It just affects Locked workflows.

Gvaihir:
> > WIP: Default destination for the file uploads is in the execution container, not the project. Trying to figure out the right way to fix, options may be:
> > 
> > 1. Make a frag applet aware of the intermediate outputs and push all the outputs to the container of the parent execution. This looks like the most viable option requiring the least modifications.
> > 2. For the non-top level workflows called by a frag exec - register intermediate outputs as workflow outputs.
> > 3. For the non-top level workflows called by a frag exec - create a collector applet that registers all intermediate outputs and pushes them to the top workflow
> 
> Would it be possible to list in the description in what scenarios the propagation to the top-level (parent) workflow is needed (and being added), e.g. when a workflow is run from an applet but not when a regular sub-workflow is called in a stage.

Of note - this is a list of options for implementation of output propagation, as opposed to a list of when the feature will be necessary/working.
If the feature is on - all intermediate outputs will be propagated, disregarding whether the workflows was or was not wrapped. 
**If wrapped**: see the description of the ticket. This was the main reason why this ticket was created. This code solves the problem: now intermediate outputs are elevated to the parent execution.
**If nested but NOT wrapped**: the intermediate outputs are still included into the outputs of the nested workflow. Meaning the process of output propagation still happens. See one additional output in the run [after](https://staging.dnanexus.com/projects/Fy9QqgQ0yzZbg9KXKP4Jz6Yq/monitor/analysis/GBZGxp80yzZkjfPJ2p56QB6K) vs [before](https://staging.dnanexus.com/projects/Fy9QqgQ0yzZbg9KXKP4Jz6Yq/monitor/analysis/G9p76YQ0yzZVjG1X5J7x20ZB) the feature implementation. 
This results from the inconsistent behavior of how intermediate outputs were handled from the beginning prior to my implementation. Meaning, there was no mechanism to explicitly propagate intermediates form unwrapped nested workflows - they were cloned by the mechanisms of platform job execution [here](https://github.com/dnanexus/dxCompiler/blob/8839cb1d8eaf0077b4bcd03d07cabbf20776b2fb/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala#L529-L530).
This current feature explicitly propagates everything, even if there's an implicit mechanism used by the platform. My suggestion was to make it consistent and either propagate all or none (depending on the feature switch), independent on the presence of a frag applet. We discussed it with @emiloslavsky but the decision was made to disregard this inconsistency for now.
Sorry - that's really a complicated behavior to explain in a short comment

core/src/main/scala/dx/core/languages/wdl/WdlCodeGenerator.scala
Gvaihir: Sorry not sure I understand - there's no relation to tests. This comment is regarding the behavior introduced in this ticket (APPS-1175), no relation to any particular test.

compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala
Gvaihir: Done, [APPS-1230](https://jira.internal.dnanexus.com/browse/APPS-1230) created. New commit will the initial approval.

core/src/main/scala/dx/core/languages/wdl/WdlCodeGenerator.scala
commandlinegirl: Would it make it a bit clearer to future developers if it mentioned what the introduced behavior is? For example: Hack to ensure proper output names when outputs are propagated to the parent-level workflow. The field names will be qualified with their namespace names to ensure they don't conflict (or similar, it's probably not the best description).

Gvaihir: ah, I see. Ok - I'll elaborate further.

commandlinegirl (dismissed):
LGTM, before merging we'll still add a switch for this feature, correct?

core/src/main/scala/dx/core/languages/wdl/WdlCodeGenerator.scala
Gvaihir: fixed in [843e357](https://github.com/dnanexus/dxCompiler/pull/342/commits/843e357fb415bbe4814915a5cb66d96f3b18af4b)

Gvaihir:
> LGTM, before merging we'll still add a switch for this feature, correct?

yes. Thanks!

commandlinegirl (approved):

Gvaihir:
This approach has been scrapped. For more info - ask @emiloslavsky

emiloslavsky:
We're planning on implementing a DNAnexus-level switch to save job outputs to the project, which shouldn't increase the number of files being passed between jobs.

PR #344, opened 2022-05-12
APPS-1197 Hot fix default instance of a native exec wrapped in frag

Gvaihir:
Fixes default instance


compiler/src/test/resources/bugs/apps_1197_native_frag_default_unit.wdl
```diff
@@ -0,0 +1,36 @@
...
+
+workflow apps_1197_native_frag_default_unit {
+
+    # Memory
+    call apps_1197_default_isntance {}
+
+    # Conditional mem
+    Boolean a = true
+    if (a) {
+        call apps_1197_default_isntance as conditional_mem {}
```
YuxinShi0423:
```suggestion
        call apps_1197_default_instance as conditional_mem {}
```

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -139,6 +139,28 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
+      .workflow(wfId)
+      .describe()
+      .stages
+      .get
+    val stagesSysReq = stages.map { stage =>
+      stage.name -> stage.systemRequirements
+    }.toMap
+    stagesSysReq.size shouldBe 2
+    stagesSysReq("if (a)") shouldBe JsObject.empty
+    stagesSysReq("apps_1197_default_isntance") shouldBe JsObject.empty
```
YuxinShi0423:
```suggestion
    stagesSysReq("apps_1197_default_instance") shouldBe JsObject.empty
```

compiler/src/test/resources/bugs/apps_1197_native_frag_default_unit.wdl
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+
+}
+
+
+task apps_1197_default_isntance {
```
YuxinShi0423:
```suggestion
task apps_1197_default_instance {
```

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
```diff
@@ -569,7 +569,7 @@ case class WorkflowCompiler(separateOutputs: Boolean,
                   if (app.kind.getClass == classOf[ExecutableKindWfFragment]) =>
                 "do not override"
               case static: StaticInstanceType => instanceTypeDb.apply(static.req).name
-              case DefaultInstanceType if (app.kind.getClass == classOf[ExecutableKindNative]) =>
+              case DefaultInstanceType =>
```
YuxinShi0423, line 572: https://github.com/dnanexus/dxCompiler/blob/c7b3061e49f5e7f413e0481fd880a8393655a256/compiler/src/main/scala/dx/translator/wdl/CallableTranslator.scala#L684-L701

So here the frag applet should always have the same instance type as its callee, and if the frag is wrapping a native applet with `DefaultInstanceType`, its own instance type should also be preserved, right? @Gvaihir

compiler/src/test/resources/bugs/apps_1197_native_frag_default_unit.wdl
Gvaihir: fixed in [868e433](https://github.com/dnanexus/dxCompiler/pull/344/commits/868e43320edaacb2f5cedbc686437bdb6ae8f666)

compiler/src/test/scala/dx/compiler/CompilerTest.scala
Gvaihir: fixed [868e433](https://github.com/dnanexus/dxCompiler/pull/344/commits/868e43320edaacb2f5cedbc686437bdb6ae8f666)

compiler/src/test/resources/bugs/apps_1197_native_frag_default_unit.wdl
Gvaihir: fixed in [868e433](https://github.com/dnanexus/dxCompiler/pull/344/commits/868e43320edaacb2f5cedbc686437bdb6ae8f666)

Gvaihir: thanks for checking the typos!

compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala
Gvaihir: The frag applet should always run on a small default instance. What happens here is:
1. Frag is translated with the same instance type as callee.
2. In [`ApplicationCompiler`](https://github.com/dnanexus/dxCompiler/blob/7592ed8d566a31b9438032596f737af69e4788f2/compiler/src/main/scala/dx/compiler/ApplicationCompiler.scala#L171) it gets overridden by specifying a default small instance from `InstanceTypeDb`, for current project it's `mem1_ssd1_v2_x2`.
3. In the [`WorkflowCompiler`](https://github.com/dnanexus/dxCompiler/blob/9297114c9d9811611dc0523522e7377e0ca26a9e/compiler/src/main/scala/dx/compiler/WorkflowCompiler.scala#L564) the original instance type of a frag is passed to the `systemRequirements` of the native callee to override the default instance. So the original instance type of the frag is just preserved for its child process. 

It's like a trust fund - you put money there but only your (grand)children can withdraw, on a very high level lol :)

For the compiled execs it makes no sense because you don't override instances - you compile with whatever you specify (or don't specify).

YuxinShi0423: Thanks for the explanation! The metaphor is awesome :D

scripts/run_multiuser_tests.py
```diff
@@ -214,6 +214,8 @@ def test_global_wf_from_wdl():
         ))
         raise
 
+
+"""Commented out in APPS-1197 because of the fund limits on Bob's account
```
YuxinShi0423, line 218: or maybe just remove the test from integration and add a separate one?
https://github.com/dnanexus/dxCompiler/blob/ed246aacb53bc955b7506f0718c377f0cff1abe7/.github/workflows/integration.yml#L67-L73

YuxinShi0423 (approved):
LGTM. Thanks 👍

PR #352, opened 2022-06-17
APPS-1222 Fix: optional types with default None

Gvaihir:
Fixed optional types coercion when the value is defaulted to `None`. Was thrown when compiling in the unlocked mode.


compiler/src/main/scala/dx/compiler/Compiler.scala
```diff
@@ -60,7 +60,6 @@ case class Compiler(extras: Option[Extras],
                     archive: Boolean,
                     force: Boolean,
                     leaveWorkflowsOpen: Boolean,
-                    locked: Boolean,
```
Gvaihir, line 63: Here and below - `Compiler` does not use `locked` arg in any methods

commandlinegirl (dismissed):
LGTM. Is [this](https://github.com/dnanexus/dxCompiler/blob/ed246aacb53bc955b7506f0718c377f0cff1abe7/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala#L378) a place where we also should make an update? 

Friendly reminder: when making cosmetic changes (commits) to the PR please remove the `integration` label so that jobs are not run for these updates. Thanks :)

EDIT: sorry for intervening in the tests but once I've noticed the unit tests failed, I canceled integration tests so as not to trigger more jobs. Consider removing the label and adding it once the unit tests pass, thanks!

Gvaihir:
> LGTM. Is [this](https://github.com/dnanexus/dxCompiler/blob/ed246aacb53bc955b7506f0718c377f0cff1abe7/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala#L378) a place where we also should make an update?
> 
No, this handles when value is not optional. Both `ValueNone` and `ValueNull` exist in the `wdlTools.typedAbstractSyntax` and possibly handle different cases. And your reference is actually a good example of that.

> Friendly reminder: when making cosmetic changes (commits) to the PR please remove the `integration` label so that jobs are not run for these updates. Thanks :)

I didn't consider this a cosmetic change. It actually affects somewhat fundamental coercions, so I did want to make sure there's no regression. I will keep integration running, when unit tests pass (they were passing locally though. EDIT: forgot to delete one redundant test, [8d53464](https://github.com/dnanexus/dxCompiler/pull/352/commits/8d534646654ccbdf857b2a060c44f57f3b5fc827))

> EDIT: sorry for intervening in the tests but once I've noticed the unit tests failed, I canceled integration tests so as not to trigger more jobs. Consider removing the label and adding it once the unit tests pass, thanks!


Sure, thanks!

commandlinegirl:
> > LGTM. Is [this](https://github.com/dnanexus/dxCompiler/blob/ed246aacb53bc955b7506f0718c377f0cff1abe7/executorWdl/src/main/scala/dx/executor/wdl/WdlWorkflowExecutor.scala#L378) a place where we also should make an update?
> 
> No, this handles when value is not optional. Both `ValueNone` and `ValueNull` exist in the `wdlTools.typedAbstractSyntax` and possibly handle different cases. And your reference is actually a good example of that.
> 
> > Friendly reminder: when making cosmetic changes (commits) to the PR please remove the `integration` label so that jobs are not run for these updates. Thanks :)
> 
> I didn't consider this a cosmetic change. It actually affects somewhat fundamental coercions, so I did want to make sure there's no regression. I will keep integration running, when unit tests pass (they were passing locally though. EDIT: forgot to delete one redundant test, [8d53464](https://github.com/dnanexus/dxCompiler/pull/352/commits/8d534646654ccbdf857b2a060c44f57f3b5fc827))

> > EDIT: sorry for intervening in the tests but once I've noticed the unit tests failed, I canceled integration tests so as not to trigger more jobs. Consider removing the label and adding it once the unit tests pass, thanks!
> 
> Sure, thanks!

Thanks :) Agreed re running this PR through these tests. In my comment I meant subsequent, potentially cosmetic commits in the PR, involving comments, documentation, etc :)

commandlinegirl (approved):

PR #355, opened 2022-07-20
APPS-1282 fix optional source input merging

YuxinShi0423:
When using "linkMerge": "merge_flattened":
https://www.commonwl.org/v1.2/Workflow.html#Merging_multiple_inbound_data_links
> Source parameters which are arrays are concatenated. Source parameters which are single element types are appended as single elements.

This PR is to fix the merging implementation when the source value is an optional array.


test/cwl_conformance/workflows/count-lines12-wf.cwl.json
```diff
@@ -15,6 +15,7 @@
     {
       "id": "file2",
       "type": [
+        "null",
```
YuxinShi0423, line 18: Tested 
https://staging.dnanexus.com/projects/Fy9QqgQ0yzZbg9KXKP4Jz6Yq/monitor/analysis/GFFZpzQ0yzZvZvYg0VZfyFJG

executorCwl/src/main/scala/dx/executor/cwl/CwlWorkflowExecutor.scala
```diff
@@ -372,7 +372,7 @@ case class CwlWorkflowExecutor(workflow: Workflow,
             if (stepInput.linkMerge.nonEmpty || stepInput.pickValue.nonEmpty) {
               val mergedValues = (sourceValues, stepInput.linkMerge) match {
                 case (_, Some(LinkMergeMethod.MergeFlattened)) =>
-                  sourceValues.flatMap {
+                  sourceValues.map(s => (CwlOptional.unwrapOptional(s._1), s._2)).flatMap {
```
YuxinShi0423, line 375: `sourceValues` could be an array of optional items. So without `unwrapOptional` those sources with type `CwlOptional(t)` will not match the `case (array: CwlArray)` below.

Gvaihir (dismissed):

commandlinegirl (dismissed):
LGTM. The only thing I'd suggest is clearer (a bit more descriptive) release notes

commandlinegirl (dismissed):
LGTM, thanks! I think the conformance tests don't need to be rerun for the last commit.

Gvaihir:
> LGTM, thanks! I think the conformance tests don't need to be rerun for the last commit.

Hm, I did delete the label

Gvaihir (approved):

commandlinegirl (approved):

PR #363, opened 2022-07-25
Add "chr"

kyclark:
Missing "chr" prefix so this fails.


Gvaihir:
Depending on the used annotation, chromosome format may or may not have `chr` prefix. Suggested changes do not add more generalization or fix any problem. Please use the provided code with the notion that it may need to be tailored for each specific use case.

PR #366, opened 2022-08-02
APPS-1225 Update add file object metadata dxc

Gvaihir:
PR for propagation of metadata attr of File type to dxCompiler from `cwlScala`


core/src/main/scala/dx/core/ir/Value.scala
```diff
@@ -32,14 +32,17 @@ object Value {
     * @param checksum the file checksum
     * @param size the file size
     * @param secondaryFiles additional files that must accompany this file
+    * @param format File format according to the language specification
```
YuxinShi0423, line 35: Nice catch! Thanks :)

YuxinShi0423 (approved):
Awesome work! Thank you so much for completing it :D

PR #372, opened 2022-08-19
APPS-1303 delayWorkspaceDestruction preserve

Gvaihir:
Fix preserving delayWorkspaceDestruction throughout execution tree

(feat) ExecutableTree.scala - generalized exec tree parsing
(feat) CompilerTest.scala - check for all stages to have identical delayWorkspaceDestruction setting


Gvaihir:
Reported bug could not be reproduced. Added unit test to monitor that all layers of the workflow have `delayWorkspaceDestruction: true` when specified with extras.

commandlinegirl (approved):

PR #380, opened 2022-08-31
APPS-1389 dxCint test failures

JanProvaznik:


dxcint/tests/fixtures/resources/mock_category/mock_1_input.json
```diff
@@ -1,3 +1,3 @@
 {
-  "mock_1.in_1": "Hello World!"
+  "stage-common.in_1": "Hello World!"
```
Gvaihir: Pls revert because:
1. We don't have common or output stages in out integration tests (unless those stages are explicitly tested) - waste of money
2. If a workflow is locked, it accepts only top level inputs (see [`locked`](https://github.com/dnanexus/dxCompiler/blob/develop/doc/ExpertOptions.md#compiling-workflow)).

dxcint/tests/conftest.py
```diff
@@ -19,10 +20,12 @@ def fixtures_dir():
 
 @pytest.fixture(scope="session")
 def context_init(root_dir):
+    timestamp = str(datetime.datetime.now()) \
+        .replace(' ','T').replace(':','-').split('.')[0]
     yield Context(
         project="dxCompiler_playground",
         repo_root=root_dir,
-        folder="/gvaihir/dxcint_testing",
+        folder=f"/dxcint_testing/test_{timestamp}",
```
Gvaihir: thanks for reorganizing!

dxcint/dxcint/Terraform.py
```diff
@@ -98,33 +98,40 @@ def _make_prerequisites(self, language: str) -> str:
...
+        def find_asset_obj(self: Terraform, asset_name: str, zero_ok: bool) -> str:
+            return dxpy.search.find_one_data_object(
+                classname="record",
+                project=self._context.project_id,
+                name=asset_name,
+                folder=self._context.platform_build_dir,
+                more_ok=False,
+                zero_ok=zero_ok
+            )
+
```
Gvaihir: I would recommend avoiding nested declaration of functions. - the less nested code you have the better it is for the readability. If you do that to keep the interface of a nested function simpler (i.e. by using already declared vars in the outer scope) - you risk to lose a track of the state of those vars during the execution of the outer function. 
Since we don't need `zero_ok` (see my other comment) - I would revert.

dxcint/dxcint/Terraform.py
```diff
@@ -98,33 +98,40 @@ def _make_prerequisites(self, language: str) -> str:
...
         asset_name = f"dx{language}rt"
+        # skip building, if asset is already present
+        asset_obj = find_asset_obj(self, asset_name, zero_ok=True)
+        if asset_obj is not None:
+            return asset_obj.get("id")
```
Gvaihir: Nice optimization - but we do wanna rebuild assets for every test run, because we are testing the code changes. And the assets represent those executors of the workflows that are the part of dxCompiler. If you rely on the previous build - you won't be able to test the code changes.
Of note - previous testing script did have some options to skip rebuilding which IMHO made UX more complex and I never used them. Please revert

Gvaihir (changes requested):
Thanks for the changes!
Pls revert where I requested

dxcint/tests/fixtures/resources/mock_category/mock_1_input.json
JanProvaznik: There is this failure when I revert:
FAILED dxcint/tests/test_RegisteredTest.py::test_run_wdl - dxpy.exceptions.DXError: DXWorkflow: the given stage identifier mock_1 could not be found as a stage ID nor as a stage name

When I remove the "mock_1." and add -locked to RegisteredTest:74 it works, but without the -locked it does not, is that the intended behaviour?

Gvaihir: ok, just keep it as `stage-common.in_1` - locked or not, will be added in case classes

dxcint/tests/fixtures/resources/mock_category/mock_1_input.json
```diff
@@ -1,3 +1,3 @@
 {
-  "mock_1.in_1": "Hello World!"
+  "in_1": "Hello World!"
```
Gvaihir, line 2:
```suggestion
  "stage-common.in_1": "Hello World!"
```

Gvaihir: reverted back, agree on your point @JanProvaznik

Gvaihir (approved):
Good job!

PR #383, opened 2022-09-07
APPS-1397 dxcint black and flakes8 pre-commit hook

JanProvaznik:
Please review after APPS-1378 is merged to show only the diff from that and not from develop now.


Gvaihir:
For the future pls create new branches off of `develop`

Gvaihir (dismissed):
Nice! I'll add `.editorconfig` to the `dxcint` root

Gvaihir (approved):

PR #397, opened 2022-10-06
APPS-1385 Update false dependency

Gvaihir:
(fix) Internals.md - docs


commandlinegirl (approved):

doc/Internals.md
```diff
@@ -441,6 +441,26 @@ complete, and returns an array of integers.
...
+This means all downstream calls depending on `z` will depend on `mul`. This is prone to decrease the parallelism of some workflows 
+due to this unwanted dependency on `mul`. This should be taken in consideration when designing the workflows.  
+Another situation possible when `z` itself is **not** dependent on previous tasks, like below:
+```wdl
+    ...
+    Int z = 1+2
+    call mul { input: a=z, b=5 }
+    ...
+```
+Here, we recommend to move such declarations to the input section of the top level workflow. This will move
```
emiloslavsky, line 461: does this solution only apply to the top-level workfow?
Can you add the rest of the toy workflow to illustrate your suggested fix?

Also, adding z to the input section would also unnecessarily expand the interface of the workflow, which is not always desired.

Gvaihir: > does this solution only apply to the top-level workfow?

Does not matter - if there's an expression in the workflow inputs - there will be a common applet which will serve as a common dependency. This is a more straightforward solution than adding a "dummy" applet. 


> Can you add the rest of the toy workflow to illustrate your suggested fix? 

I based my tests on your investigation in the ticket. I see no reason to add a full WF to the readme as it may dilute the message.

> Also, adding z to the input section would also unnecessarily expand the interface of the workflow, which is not always desired.

I agree and posted my long term suggestions in the comments to the APPS-1385.

PR #408, opened 2022-12-13
APPS-1491 Update dxda version bump

Gvaihir:
(fix) bundled_dependencies.json - current deployment 
(fix) dxda.json - dxcint deployment for the future

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## in develop
+
+* `dxda` update to `v0.5.10` which adds a retry for file download 
+
 ## 2.10.6 2022-10-25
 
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## in develop
+
+* `dxda` update to `v0.5.10` which adds a retry for file download
```
commandlinegirl: Nit: I'd only maybe make it even more clear which error it fixes so the client knows immediately it was "their" issue, e.g. which adds a retry for file download fixing the "context canceled" error (or similar)

commandlinegirl (dismissed):

RELEASE_NOTES.md
Gvaihir: Fixed with more info assuming "their" means "originated form dxda".

commandlinegirl (approved):
Thank you!

PR #422, opened 2023-02-24
APPS-1676 HOTFIX - update ofh and other Azure instance types to v2

Gvaihir:
(fix) dxapp.json - updated instances for azure


commandlinegirl (approved):
Thanks for the quick fix!

PR #425, opened 2023-03-04
APPS-1676 HOTFIX ofh instance

Gvaihir:
Just a few updates to the cloning app that I forgot to merge. This is just to keep track - since the app was already built and functions as a part of APPS-1676

### DIFF
--- a/scripts/dxcompiler_copy/dxapp.json
+++ b/scripts/dxcompiler_copy/dxapp.json
@@ -3,7 +3,7 @@
   "title": "Copy File",
   "summary": "Copy a file to a different region",
   "dxapi": "1.0.0",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "billTo": "org-dnanexus_apps",
   "developers": [
     "org-dnanexus_apps"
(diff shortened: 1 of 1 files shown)


commandlinegirl (approved):
LGTM but is this instance available in the OFH region?

BTW, in our instance type database do we take into account the fact the some regions have a different set of ITs available than other regions?

Gvaihir:
> BTW, in our instance type database do we take into account the fact the some regions have a different set of ITs available than other regions?

Not sure what exactly you are asking about but:
* In the component [InstanceTypeDb](https://github.com/dnanexus/dxScala/blob/2dabaab10bcfbdf3433f6e730c9215a406cce686/api/src/main/scala/dx/api/DxInstanceType.scala#L528) - nothing is hardcoded.
* In the `dxwdl_copy` [app spec](https://github.com/dnanexus/dxCompiler/blob/9c35599d513eb0c764a1dc1d576fcbf21dd49c7e/scripts/dxcompiler_copy/dxapp.json#L2), all default instances are based on available instance of the region. If the instance type is not available we would be getting [this error.](https://github.com/dnanexus/dxCompiler/actions/runs/4258987371/jobs/7410773182#step:9:372)

PR #432, opened 2023-03-16
APPS-1712 Post release 2.11.0

Gvaihir:


compiler/src/main/resources/application.conf
```diff
@@ -1,5 +1,5 @@
 dxCompiler {
-    version = "2.10.10-SNAPSHOT"
+    version = "2.11.1-SNAPSHOT"
```
r-i-v-a, line 2: version 2.11.0?

r-i-v-a (changes requested):
version 2.11.0-SNAPSHOT?

compiler/src/main/resources/application.conf
r-i-v-a: Sorry, never mind

r-i-v-a (approved):

PR #436, opened 2023-04-14
HOTFIX_openSSL_version

Gvaihir:
(fix) scala.yml - pinned dxpy version.
(fix) InstanceTypesTest.scala - removed info

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -45,7 +45,7 @@ jobs:
           cwltool=`jq -r '.execDepends.cwl[0].version' scripts/bundled_dependencies.json`
           salad=`jq -r '.execDepends.cwl[1].version' scripts/bundled_dependencies.json`
           pip3 install --upgrade pip setuptools==58.3.0
-          pip3 install termcolor dxpy cwltool==$cwltool schema-salad==$salad pyOpenSSL==22.0.0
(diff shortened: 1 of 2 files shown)


.github/workflows/scala.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           cwltool=`jq -r '.execDepends.cwl[0].version' scripts/bundled_dependencies.json`
           salad=`jq -r '.execDepends.cwl[1].version' scripts/bundled_dependencies.json`
           pip3 install --upgrade pip setuptools==58.3.0
-          pip3 install termcolor dxpy cwltool==$cwltool schema-salad==$salad pyOpenSSL==22.0.0
+          pip3 install termcolor dxpy==0.323.0 cwltool==$cwltool schema-salad==$salad pyOpenSSL==22.0.0
```
commandlinegirl, line 48: Since the latest version is 0.345.0, can we pin it to that version?

Gvaihir: This is the reason of the hotfix: [here](https://github.com/dnanexus/dxCompiler/actions/runs/4702499161/jobs/8339837073#step:6:83). Latest version of dxpy is incompatible with the pinned version of OpenSSL. Later versions of OpenSSL had a bug - something with parsing Ubuntu distr name.

commandlinegirl: Oh, got it

commandlinegirl (approved):

PR #444, opened 2023-05-27
APPS-1881 file describe optimization

Gvaihir:
Main changes to optimize for file/describe API calls.
Depends on https://github.com/dnanexus/dxScala/pull/119


executorCommon/src/main/scala/dx/executor/JobMeta.scala
```diff
@@ -403,6 +403,35 @@ abstract class JobMeta(val workerPaths: DxWorkerPaths,
...
+
+    val notClosed = {
+      if (verifyClosed) {
+        logger.trace(s"Checking that all files are closed")
+        dxFiles.filterNot(_.describe().state == DxState.Closed)
+      } else {
+        Vector.empty
+      }
+    }
+    if (notClosed.nonEmpty) {
```
commandlinegirl, line 426: If the files are not closed (e.g in the process of closing) would it make sense to retry or wait before throwing an error? Or we expect a non-closed file to remain open/closing at this point anyway?

Gvaihir: dxCompiler `-waitOnUpload` covers that. Previous behavior was that if `-waitOnUpload` was not active - no checks were done to the delocalized files at all, and `file-xxx/describe` was called only to get a URI of the file on `dx://` I basically preserve this behavior.

commandlinegirl: OK, sg, thanks!

commandlinegirl (dismissed):
LGTM, thank you!

commandlinegirl (approved):
LGTM, thank you!

PR #447, opened 2023-06-06
APPS-1857 tar_io_doc

Gvaihir:
(feat) Tips.md - tar advice

### DIFF
--- a/doc/Tips.md
+++ b/doc/Tips.md
@@ -123,9 +123,9 @@ workflow wf_tar {
 Since the `File` type declaration is represented by a string value, please follow the official [WDL documentation](https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#strings) 
 regarding the format and allowed characters.  
 ### Examples
-**Scenario 1.** Stage A creates `N` files to be used only by Stage B, and there is no need for a separate archiving task.
-In this case the archiving can be done in the Stage A `task` section, and un-archiving should be done in the `task` 
-section of the Stage B.
+**Scenario 1.** Task A creates `N` files to be used only by Task B, and there is no need for a separate archiving task.
+In this case the archiving can be done in the Task A `task` section, and un-archiving should be done in the `task` 
+section of the Task B.
 ```wdl
 task task_a {
 ...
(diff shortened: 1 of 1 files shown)


doc/Tips.md
```diff
@@ -75,3 +75,95 @@ task make_error {
 
 (including the hostname in the error message allows the message to have the root
 job ID, facilitating debugging in complex workflows that may be deeply nested.)
+
+
+## Large number of input/output files
+When passing a large number of files from stage to stage, we recommend doing that in the form of a `tar.gz` archive to
```
commandlinegirl: Can you provide a ballpark amount of files we mean, e.g. 

```
... a large number of files, i.e. larger than 1000, from ...
```

?

doc/Tips.md
```diff
@@ -75,3 +75,183 @@ task make_error {
...
+          input:
+          single_file = s
+        }
+    }
+...
+```
+Since the `File` type declaration is represented by a string value, please follow the official [WDL documentation](https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#strings) 
+regarding the format and allowed characters.  
+### Examples
+**Scenario 1.** Stage A creates `N` files to be used only by Stage B, and there is no need for a separate archiving task.
```
commandlinegirl: Here you refer to `Stage A` and `Stage B` but in the code there are tasks, so I'd suggest to change it to Task A/B.

doc/Tips.md
```diff
@@ -75,3 +75,183 @@ task make_error {
...
+}
+
+workflow wf_tar {
+...
+    call task_a {...}
+    call task_b {input: input_tarball=task_a.tarball}
+...
+}
+```
+**Scenario 2.** Stage A creates `N` files to be used by Stage B, C, ... m, and it makes sense to refactor the archiving
```
commandlinegirl:
```suggestion
**Scenario 2.** Stage A creates `N` files to be used by multiple tasks B, C, ...M, here represented by `task_b_c_through_m`, by and it makes sense to refactor the archiving 
```

commandlinegirl (changes requested):
A couple of minor requests, this is useful, thank you!

doc/Tips.md
Gvaihir: Hm, that was addressed in [306022e](https://github.com/dnanexus/dxCompiler/pull/447/commits/306022eede1280c6c5b79076a3c7d631cb69fa6c)

doc/Tips.md
Gvaihir: Fixed in [04792e0](https://github.com/dnanexus/dxCompiler/pull/447/commits/04792e0448098576d9227896dc790ad81b5028dd)

commandlinegirl (approved):
LGTM, thank you!

doc/Tips.md
commandlinegirl: True, it was a stale comment from when I started the review. Resolving.

PR #473, opened 2024-09-30
APPS-2624 Put awscliv2 ecr get-login-password into retry loop

r-i-v-a:
APPS-2624 Put awscliv2 ecr get-login-password into retry loop

### DIFF
--- a/compiler/src/main/resources/templates/ecr_docker_preamble.ssp
+++ b/compiler/src/main/resources/templates/ecr_docker_preamble.ssp
@@ -35,20 +35,20 @@
 
         # get the password from the aws client
         password=${bashDollar}(awscliv2 ecr get-login-password --region ${bashDollar}{AWS_REGION})
-        if [[ $? -ne 0 ]]; then
+        if [[ ${bashDollar}? -ne 0 ]]; then
             echo "Could not get password from AWS"
-            continue
+        else
+            # TODO: use credential helper instead https://docs.docker.com/engine/reference/commandline/login/
+            echo ${bashDollar}{password} | docker login ${bashDollar}{DOCKER_REGISTRY} --username AWS --password-stdin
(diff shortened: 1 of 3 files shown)


aogorodnikov-dna (dismissed):
Approval for the code changes - don't know if it will solve the API throttling at the current backoff strategy

compiler/src/main/resources/templates/ecr_docker_preamble.ssp
```diff
@@ -27,14 +27,15 @@
...
-    password=${bashDollar}(awscliv2 ecr get-login-password --region ${bashDollar}{AWS_REGION})
-
     # Log into docker. Retry several times.
     i=0
     while [[ ${bashDollar}i -le 5 ]]; do
         # don't exit if docker command fails - retry instead
         set +e
+
+        # get the password from the aws client
+        password=${bashDollar}(awscliv2 ecr get-login-password --region ${bashDollar}{AWS_REGION})
```
mhrvol, line 37: Shall we add something like 

`if [[ $? -ne 0 ]]; then
    echo "Could not get password from AWS"
    continue
fi`

To prevent unnecessary docker login errors?

r-i-v-a: done

r-i-v-a:
Edited, please review again @mhrvol @aogorodnikov-dna

compiler/src/main/resources/templates/ecr_docker_preamble.ssp
mhrvol: We still don't "continue" if the result value is not 0, also I am not wondering if the echo message won't be confusing as we are going to retry anyway.

mhrvol (approved):
LGTM

PR #491, opened 2025-02-12
DEVEX-2460 Ubuntu update to Noble

mhrvol:
Tested only on simple workflows, not on any advanced stuff like reorgs. 

More tests are to be added tomorrow.

### DIFF
--- a/compiler/src/main/scala/dx/translator/Extras.scala
+++ b/compiler/src/main/scala/dx/translator/Extras.scala
@@ -43,12 +43,22 @@ object ExtrasJsonProtocol extends DefaultJsonProtocol {
             timeout("*").convertTo[DxTimeout]
           }
           val headJobOnDemand = JsUtils.getOptionalBoolean(fields, DxRunSpec.HeadJobOnDemand)
+          val release =
+            JsUtils.getOptionalString(fields, DxRunSpec.release).map {
+              case release if Constants.allowedOsReleases.contains(release) => release
+              case release =>
+                throw new Exception(
+                    s"Unsupported execution environment value ${release}." +
+                      s" Supported values are ${Constants.allowedOsReleases.mkString(", ")}"
+                )
+            }
           DxRunSpec(
               fields.get(DxRunSpec.Access).map(_.convertTo[DxAccess]),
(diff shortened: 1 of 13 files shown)


r-i-v-a:
@mhrvol could you also mention it in this documentation? https://github.com/dnanexus/dxCompiler/blob/develop/doc/ExpertOptions.md#setting-dnanexus-specific-attributes-in-extrasjson

r-i-v-a (changes requested):
update documentation

compiler/src/main/scala/dx/translator/Extras.scala
```diff
@@ -263,6 +274,7 @@ case class DxRunSpec(access: Option[DxAccess],
                      executionPolicy: Option[DxExecPolicy],
                      restartableEntryPoints: Option[String],
                      timeoutPolicy: Option[DxTimeout],
+                     release: Option[String],
```
aogorodnikov-dna, line 277: Maybe default value here to avoid changes in `ExtrasTest`?

aogorodnikov-dna (approved):
All good - maybe one default `None` would make your life easier and avoid changes in tests. See more in place

compiler/src/main/scala/dx/translator/Extras.scala
mhrvol: I had the default at first, but saw that other things did not have any default (i.e. timeoutPolicy real default is 30d) - so I wanted to stick with this.

Also release=None wouldn't really work right? We would have default which could never be used.

I could make the default Some(Constants.defaultOs), but again - it's very different to other arguments.

mhrvol: On the other hand, maybe I could add the None default to more arguments?

whichever you think is preferable - I'll apply it to the code

r-i-v-a (dismissed):

compiler/src/main/scala/dx/translator/Extras.scala
mhrvol: @aogorodnikov-dna

r-i-v-a (approved):

PR #500, opened 2025-03-28
COMPUTE-1237: Update dxApi version to handle 429 response code from platform

windok:
[COMPUTE-1236](https://jira.internal.dnanexus.com/browse/COMPUTE-1237)

* Update dxApi sdk dependency to support 429 http response code from platform.
* Update dxpy to 0.393.0
* Update dev docs


executorCommon/src/main/resources/application.conf
```diff
@@ -1,5 +1,5 @@
 dxExecutorCommon {
-    version = "2.13.1-SNAPSHOT"
+    version = "2.13.1"
```
r-i-v-a: keep the version on -SNAPSHOT unless doing a release

RELEASE_NOTES.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+## 2.13.1 2025-03
```
r-i-v-a: Remove the line with "## 2.13.1 2025-03" -- keep the top header as unreleased, it will be updated when doing the next release.

r-i-v-a:
Tests passed; removing labels so they won't be re-run for minor changes (changelog format, etc.)

r-i-v-a (changes requested):
comments re: snapshot version and changelog

executorCommon/src/main/resources/application.conf
windok: Ok, got you. I forgot that release branch should not be merged.

I returned `SNAPSHOT` back

RELEASE_NOTES.md
windok: Fixed

r-i-v-a (approved):

PR #508, opened 2025-06-03
DEVEX-2580 Remove deprecated GHA workflow containing typo

r-i-v-a:
DEVEX-2580 Remove deprecated GHA workflow containing typo


YuxinShi0423 (approved):
LGTM. But not sure if removing it does not harm and if we should also change the integration2 to integration.

PR #510, opened 2025-06-03
DEVEX-2578 dxCompiler Option for Parallel Compilation of Applets [by Fulcrum Genomics]

r-i-v-a:
DEVEX-2579 For testing Fulcrum Genomics PR


scripts/clean_build.sh
```diff
@@ -1,12 +1,14 @@
 #!/bin/bash
+set -euxo pipefail
 
-TEST=${1}
+TEST=${1:-""}
```
r-i-v-a, line 4: - [x] revert script

r-i-v-a:
- [x] Add documentation about `-executableCreationParallelism <int>`

r-i-v-a:
- [x] Add changelog note

compiler/src/test/scala/dx/compiler/CompilerTest.scala
```diff
@@ -973,6 +978,14 @@ class CompilerTest extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
...
     )
 
+    if (desc1.access == expected2Access && desc2.access == expected1Access) {
+      fail("Compiled applet order is reversed")
+    }
+
+    desc1.access shouldBe expected1Access
+
+    desc2.access shouldBe expected2Access
+
```
YuxinShi0423, line 988: I dont get it why the access order could be wrong. The parallelized compilation is done on callable, and the network access is embedded in the callable object, so it would go with the callable and should not be affected by this new setting. @r-i-v-a

r-i-v-a: @YuxinShi0423 I think the test is just asserting that the new setting doesn't change the order; this was mentioned in the spec attached to the PMUX ticket.

TedBrookings: I had introduced a bug that _did_ reverse the order, but the resulting error message was unclear (it looked like the results were just wrong). I made the message for reversed order more clear and fixed the bug.

YuxinShi0423: @TedBrookings could you point me to the commit for that fix?

TedBrookings: [d56c1cb Merge branch 'for_testing_fulcrum_genomics_pr' of github.com:dnanexus/dxCompiler into for_testing_fulcrum_genomics_pr](https://github.com/dnanexus/dxCompiler/pull/510/commits/d56c1cbb78c38f2236271307fc0b8aa495dd96a2)
The changes were to Compiler.scala and ExecutableTree.scala. That said, this was fixing a bug that I had introduced earlier. I had been using Map keys instead of Vectors to track dependencies, so the order of mutually-independent executables had not been preserved correctly.

YuxinShi0423: Thanks! So the access config is just the flag we used to check if applet1 and 2 are in correct order, but itself wont not be swapped.  
I would still prefer another unit test since this one serves two purposes now. But I am okay with it. @r-i-v-a

YuxinShi0423 (approved):
I got no more question. Thanks! LGTM

PR #523, opened 2025-09-04
DEVEX-2566: Add support for OCI Ashburn region

windok:
- https://jira.internal.dnanexus.com/browse/DEVEX-2566
- Created platform `dxCompiler_Ashburn` projects in [staging](https://staging.dnanexus.com/panx/projects/J2vp5GV64X35QPz2KGYjY11y/data/) and [prod](https://platform.dnanexus.com/panx/projects/J2vkJvV6f651yjK864J6K894/data/) env

### DIFF
--- a/scripts/dxcompiler_copy/dxapp.json
+++ b/scripts/dxcompiler_copy/dxapp.json
@@ -3,7 +3,7 @@
   "title": "Copy File",
   "summary": "Copy a file to a different region",
   "dxapi": "1.0.0",
-  "version": "0.0.8",
+  "version": "0.1.0",
   "billTo": "org-dnanexus_apps",
   "developers": [
     "org-dnanexus_apps"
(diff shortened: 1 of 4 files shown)


scripts/dxcompiler_copy/dxapp.json
```diff
@@ -3,7 +3,7 @@
   "title": "Copy File",
   "summary": "Copy a file to a different region",
   "dxapi": "1.0.0",
-  "version": "0.0.7",
+  "version": "0.0.8",
```
jsitarova-dnanexus: For other apps we increased minor version instead of patch, not sure if dxCompiler has other rules

windok: I also don't sure, for now there were no patch version changes, there is still version `0.0.x`

r-i-v-a: I think this is just some internal app, but we can increase the minor version to match what is done with other apps.

r-i-v-a (commented):
can increase minor rather than patch version

r-i-v-a (approved):

PR #526, opened 2025-09-25
DEVEX-2566: Add automation script to build and publish Copy File app

windok:
Automate building and publishing `Copy File` that is required for copying dxCompiler asset when new region is added.

`Makefile` was removed because it's not used anywhere and not referenced from docs. Also, in `Makefile` old app folder name is mentioned, currently app is stored in `dxcompiler_copy` folder.


jsitarova-dnanexus (approved):
LG

PR #527, opened 2025-10-06
DEVEX-2566, DEVEX-2604: Post release 2.15.0

windok:


compiler/src/main/resources/application.conf
```diff
@@ -1,5 +1,5 @@
 dxCompiler {
-    version = "2.14.1-SNAPSHOT"
+    version = "2.15.1-SNAPSHOT"
```
r-i-v-a, line 2: 2.15.0-SNAPSHOT

r-i-v-a (changes requested):
new version: 2.15.0-SNAPSHOT

windok:
I followed instruction to setup next snapshot version by upgrading patch version based on current version which is already `2.15.0`. https://github.com/dnanexus/dxCompiler/blob/develop/doc/Developing.md?plain=1#L240

Also, previously was version `2.14.0` and snapshot version was with changed patch version `2.14.1-SNAPSHOT`.

r-i-v-a:
> I followed instruction to setup next snapshot version by upgrading patch version based on current version which is already `2.15.0`. https://github.com/dnanexus/dxCompiler/blob/develop/doc/Developing.md?plain=1#L240
> 
> Also, previously was version `2.14.0` and snapshot version was with changed patch version `2.14.1-SNAPSHOT`.

I see, 2.15.0 is the version being released just now. Sorry!

r-i-v-a (approved):

PR #531, opened 2026-03-23
Add support for authenticated URLs

bvaisvil:
@XingGemini 

**Please note an LLM was used to aid in the generation of the code for this PR. I have thoroughly reviewed the changes along with (@Ben-Habermeyer and @artgramlich). But we are not scala engineers.**

## Summary

This PR adds support for authenticated HTTP imports via a new `AuthenticatedHttpFileAccessProtocol`. Users can now import private WDL/CWL files from authenticated HTTPS sources (e.g. GitHub private repos) by setting the `WDL_IMPORT_TOKENS` environment variable. 

For example: `raw.githubusercontent.com:ghp_abc;gitlab.com:glpat_xyz`

Being able to import from authenticated sources enables greater flexibility in creating WDL modules that can be re-used and imported from another WDL.

## Configuration

Set `WDL_IMPORT_TOKENS` as a semicolon-separated list:

    WDL_IMPORT_TOKENS="raw.githubusercontent.com:ghp_abc;gitlab.com:glpat_xyz"

Tokens are only sent to the matching domain; unconfigured domains
receive no `Authorization` header (same behavior as before).

No breaking changes. When `WDL_IMPORT_TOKENS` is unset, behavior is
identical to the previous `HttpFileAccessProtocol` (no auth headers).

## Testing

- Added unit tests in
  `AuthenticatedHttpFileAccessProtocolTest` covering token parsing, per-domain dispatch, case-insensitive matching, and path resolution with token propagation.
- Some unit tests unrelated to my changes fail because of access to certain dnanexus staging environments and projects.
- compiled a WDL that references a workflow on a private repo:
image
- Compiled WDL was uploaded to the platform, executed, and completed successfully.

 

## Some notes
- `AuthenticatedHttpFileSource*` live in `core.source.main.scala.dx.core.io`, whereas `FileSource` lives in dxcommon. This is where these classes probably should live, but for demonstration purposes they are here. If the general outline is considered acceptable, I can make a PR into https://github.com/dnanexus/dxScala to add our new FileSource there.
- `initCommon` constructs `FileSourceResolver` now instead of using `create`.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,3 @@
+java temurin-11.0.21+9
+sbt 1.9.7
+python 3.11.6
--- a/compiler/src/main/scala/dxCompiler/Main.scala
+++ b/compiler/src/main/scala/dxCompiler/Main.scala
@@ -14,7 +14,8 @@ import dx.core.languages.wdl.WdlOptions
 import dx.dxni.DxNativeInterface
 import dx.translator.{Extras, TranslatorFactory}
 import dx.util.protocols.DxFileAccessProtocol
-import dx.util.{Enum, FileSourceResolver, FileUtils, Logger, TraceLevel}
+import dx.util.{Enum, FileAccessProtocol, FileSourceResolver, FileUtils, LocalFileAccessProtocol, Logger, TraceLevel}
+import dx.core.io.AuthenticatedHttpFileAccessProtocol
 import spray.json.{JsNull, JsValue}
 import wdlTools.types.TypeCheckingRegime
 
@@ -67,17 +68,25 @@ object Main {
     * - creates a FileSourceResolver that looks for local files in any configured -imports
     *   directories and has a DxFileAccessProtocol
     * - initializes a Logger
+    * - configures authenticated HTTP imports if WDL_IMPORT_TOKEN is set
     * @param options parsed options
     * @return (FileSourceResolver, Logger)
     */
   private def initCommon(options: Options): (FileSourceResolver, Logger) = {
     val logger = initLogger(options)
     val imports: Vector[Path] = options.getList[Path]("imports")
-    val fileResolver = FileSourceResolver.create(
-        imports,
-        Vector(DxFileAccessProtocol()),
-        logger
+    
+    // Create authenticated HTTP protocol for importing from private repositories
+    val httpProtocol = AuthenticatedHttpFileAccessProtocol.fromEnvironment(logger)
+    
+    // Build protocol list - order matters, first matching protocol wins
+    val protocols: Vector[FileAccessProtocol] = Vector(
+      LocalFileAccessProtocol(imports, logger),
+      httpProtocol,
+      DxFileAccessProtocol()
     )
+    
+    val fileResolver = FileSourceResolver(protocols)
     FileSourceResolver.set(fileResolver)
     (fileResolver, logger)
   }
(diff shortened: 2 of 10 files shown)


.tool-versions
```diff
@@ -0,0 +1,3 @@
+java temurin-11.0.21+9
```
Ben-Habermeyer, line 1: nit: DX may ask to revert this however I personally appreciate it

core/src/main/scala/dx/core/io/AuthenticatedHttpFileAccessProtocol.scala
```diff
@@ -0,0 +1,104 @@
...
+    encoding: Charset = FileUtils.DefaultEncoding,
+    logger: Logger = Logger.Quiet
+) extends FileAccessProtocol {
+
+  override val schemes: Vector[String] = Vector(FileUtils.HttpScheme, FileUtils.HttpsScheme)
+  override val supportsDirectories: Boolean = true
+
+  /**
+    * Determines if authentication should be used for the given URI.
+    * Only returns true if a token is configured AND the domain is in the allowed list.
```
Ben-Habermeyer: checking my understanding here, if the import belongs to one of the allowed domains AND a token is set in the env then this protocol will be used. Is this done on an import-by-import basis?

What happens if we have imports from multiple sources? e.g.
```
import local.wdl
import http://raw.githubusercontent.com/remote/public.wdl
import https://github.com/company/repo/private.wdl
```

Ben-Habermeyer (commented):
looks pretty good to me, though I admit my scala knowledge is limited. I provided a pointer to their dev docs and have one question about mixed imports

compiler/src/main/scala/dxCompiler/Main.scala
```diff
@@ -67,17 +68,25 @@ object Main {
...
     * @return (FileSourceResolver, Logger)
     */
   private def initCommon(options: Options): (FileSourceResolver, Logger) = {
     val logger = initLogger(options)
     val imports: Vector[Path] = options.getList[Path]("imports")
-    val fileResolver = FileSourceResolver.create(
-        imports,
-        Vector(DxFileAccessProtocol()),
-        logger
+
```
artgramlich: I think this is right and the order look like it will be the same (imports, now http then dx file access)

Here's the code for LocalFileAccessProtocol and FileSourceResolver
https://github.com/dnanexus/dxScala/blob/8b056342eb52ebdb800b63d2c6179a4d2e6ed77f/common/src/main/scala/dx/util/FileSource.scala#L512
https://github.com/dnanexus/dxScala/blob/8b056342eb52ebdb800b63d2c6179a4d2e6ed77f/common/src/main/scala/dx/util/FileSource.scala#L736

core/src/main/scala/dx/core/io/AuthenticatedHttpFileAccessProtocol.scala
artgramlich: If I'm reading this correctly, it does it per source in the resolver (above). Likely worth just trying it just to make sure but it looks like that's what it's doing.

bvaisvil: I've changed this now to explicit, that specific tokens are only set to specific domains.

bvaisvil:
Closing since https://github.com/dnanexus/dxCompiler/pull/538 merged and https://github.com/dnanexus/dxCompiler/releases/tag/2.17.0 is released

